PR #114, opened 2018-04-09
MYST-496 connectTimeouts increased on connect

shroomist:
default timeout is 0, no timeout


src/libraries/api/tequilapi.js
```diff
@@ -24,27 +25,29 @@ export default function Constructor (teqAddr = 'http://127.0.0.1:4050') {
...
+        axioAdapter.put(conPath, {
+          consumerId: consumerId,
+          providerId: providerId
+        }, {timeout})
+      },
       disconnect: async () => axioAdapter.delete(conPath),
       status: async () => axioAdapter.get(conPath),
-      ip: async () => {
-        const res = await axioAdapter.get(conPath + '/ip')
+      ip: async (timeout = DEFAULT_TIMEOUT) => {
```
Waldz: I see usecase very strange:
- Configure default `teq = new Tequilapi('http://127.0.0.1:4050', 1000)`
- But `teq.ip()` fallbacks to default default value 5000

src/libraries/api/tequilapi.js
```diff
@@ -24,27 +25,29 @@ export default function Constructor (teqAddr = 'http://127.0.0.1:4050') {
...
     },
     connection: {
-      connect: async ({consumerId, providerId}) => axioAdapter.put(conPath, {
-        consumerId: consumerId,
-        providerId: providerId
-      }),
+      connect: async ({consumerId, providerId}, timeout = CONNECTION_TIMEOUT) => {
+        axioAdapter.put(conPath, {
+          consumerId: consumerId,
+          providerId: providerId
```
donce: `{consumerId, providerId}`

src/libraries/api/tequilapi.js
```diff
@@ -24,27 +25,29 @@ export default function Constructor (teqAddr = 'http://127.0.0.1:4050') {
       list: async () => axioAdapter.get(propPath)
     },
     connection: {
-      connect: async ({consumerId, providerId}) => axioAdapter.put(conPath, {
-        consumerId: consumerId,
-        providerId: providerId
-      }),
+      connect: async ({consumerId, providerId}, timeout = CONNECTION_TIMEOUT) => {
```
Waldz: Why it's not 3 parameters?

src/libraries/api/tequilapi.js
Waldz: Yeah, it should just 3 simple parameters?

shroomist:
comments fixed. @donce

src/renderer/store/modules/main.js
```diff
@@ -50,8 +50,15 @@ const mutations = {
...
-    state.errorMessage = err.response ? err.response.data.message : err.message
     state.showError = true
+    if (err && err.response && err.response.data && err.response.data.message) {
+      state.errorMessage = err.response.data.message
+      return
+    } else if (err.message) {
+      state.errorMessage = err.message
+      return
+    }
+    state.errorMessage = 'Unknown error'
```
donce: Can you add test for this case?

src/renderer/store/modules/main.js
```diff
@@ -50,8 +50,15 @@ const mutations = {
     state.newUser = true
   },
   [type.SHOW_ERROR] (state, err) {
-    state.errorMessage = err.response ? err.response.data.message : err.message
     state.showError = true
+    if (err && err.response && err.response.data && err.response.data.message) {
+      state.errorMessage = err.response.data.message
+      return
+    } else if (err.message) {
```
donce: `else` is not needed if you're returning in first `if`

src/libraries/api/tequilapi.js
```diff
@@ -6,45 +6,49 @@ const conPath = '/connection'
...
+        return axioAdapter.put(conPath, {
+          consumerId: consumerId,
+          providerId: providerId
+        }, null, timeout ? {timeout} : undefined)
+      },
       disconnect: async () => axioAdapter.delete(conPath),
       status: async () => axioAdapter.get(conPath),
-      ip: async () => {
-        const res = await axioAdapter.get(conPath + '/ip')
+      ip: async (timeout) => {
```
donce: It's not obvious from function signature that `timeout` is optional - maybe `(timeout = null)`.

src/libraries/api/tequilapi.js
```diff
@@ -6,45 +6,49 @@ const conPath = '/connection'
...
-  const {teqAxio, axioAdapter} = adapterFactory(teqAddr)
+export default function Constructor (teqAddr = 'http://127.0.0.1:4050', defaultTimeout = DEFAULT_TIMEOUT) {
+  const {teqAxio, axioAdapter} = adapterFactory(teqAddr, defaultTimeout)
   const api = {
     identity: {
       list: async () => axioAdapter.get(idPath),
       create: async (passphrase) => axioAdapter.post(idPath, {passphrase}),
       unlock: async ({id, passphrase}) => {
-        axioAdapter.put(idPath + '/' + id + '/unlock', {passphrase})
+        return axioAdapter.put(idPath + '/' + id + '/unlock', {passphrase})
```
donce: Why are we returning here?

src/renderer/store/modules/main.js
shroomist: done

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -1,5 +1,7 @@
 import {Tail} from 'tail'
 
+const healthCheckTimeout = 100
```
donce: Why are there 2 different healthCheckTimeout constants - one `100`, another `500` (in *monitoring.js*)?

shroomist: perhaps we don't even need a timeout here, as we are expecting first `healthcheck` to be rejected.

donce: Ok, this looks confusing - so what we're trying to optimise here with custom timeout? In case this is not rejected, we want to get response sooner when 0.5s?

Maybe we can treat all healthchecks the same - with 0.5s timeout? In that case, we can keep this value in `config.js` and use it where we need.

shroomist:
@Waldz @donce please review

src/libraries/mysterium-client/monitoring.js
```diff
@@ -36,7 +37,7 @@ class ProcessMonitoring {
 
   async _healthCheck () {
     try {
-      await this.api.healthCheck(500)
+      await this.api.healthCheck(healthCheckTimeout)
```
donce: So first healthcheck will wait 5s until timing out?

shroomist: 500ms. I didn't want to change the timings here

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -23,7 +23,7 @@ class Process {
   }
 
   start () {
-    this.tequilapi.healthCheck(100).catch(() => {
+    this.tequilapi.healthCheck().catch(() => {
```
donce: @shroomist sorry, my previous comment was about this place - we're skipping `timout` param, so we want this timeout to change to 5s intentionally?

shroomist: timeout in here really doesn't matter -- if daemon is down and nobody is listening on `:4050` this request would be rejected right away, no matter the timeout.

donce: But it might not be down - i.e. Mysterion could have been terminated with daemon running, and it won't be down when re-launching Mysterion again.

Not that important, but I'd keep `healthCheck` timeout the same everywhere, that is 0.5s.

PR #127, opened 2018-04-17
MYST-506 Refactor communication

donce:
- Fix bug: `MainCommunication#send` not working
- Use `MessageBus` instead of `ipcMain`, `ipcRenderer` and `Window` directly
- Add more message signatures to communication


src/app/communication/main-communication.js
```diff
@@ -15,19 +15,27 @@ class MainCommunication {
...
     this._send(messages.MYSTERIUM_CLIENT_LOG, data)
   }
 
-  onConnectionStatusChange (callback: (ConnectionStatusChangeData) => void) {
+  onConnectionStatusChange (callback: (ConnectionStatusChangeData) => void): void {
     this._on(messages.CONNECTION_STATUS_CHANGED, callback)
   }
 
-  _send (channel: string, data: mixed) {
+  onIdentitySet (callback: (IdentitySetData) => void) {
```
Waldz: `onIdentityReloaded`. Lets assume that it can be loaded several times during application lifetime.

src/app/communication/main-communication.js
```diff
@@ -15,19 +15,27 @@ class MainCommunication {
...
+    this._on(messages.IDENTITY_SET, callback)
+  }
+
+  onRendererLoaded (callback: () => void) {
```
Waldz: I am for non technical names, which describes business logic, what about `onMainWindowVisible`?

src/app/communication/dto.js
```diff
@@ -9,3 +9,7 @@ export type MysteriumClientLogData = {
   level: string,
   data: mixed
 }
+
+export type IdentitySetData = {
```
Waldz: It looks like `IdentityModel` from identity module.
Or `IdentityDTO` just for this event.

src/app/mysterion.js
```diff
@@ -74,15 +75,25 @@ class Mysterion {
...
+      throw new Error('Failed to open window.')
     }
 
-    this.messageBus = new MainMessageBus(this.window.send)
+    const send = this.window.send.bind(this.window)
+    this.messageBus = new MainMessageBus(send)
     this.communication = new MainCommunication(this.messageBus)
 
+    try {
+      await waitForMessage(this.communication.onRendererLoaded.bind(this.communication))
```
Waldz: This do we need to mention `this.communication` scope twice?

donce: Because that's how JS works - `this.communication.onRendererLoaded` is reference to MainCommunication.prototype.onRendererLoaded.
Good explanation here:
https://stackoverflow.com/questions/36489579/this-within-es6-class-method?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

Waldz: Not much use to have reusable helper `waitForMessage`, instead I would do:
`await this.waitForMainWindowLoad()` which wraps all hacks

donce: > Not much use to have reusable helper `waitForMessage`
There are many usages of `this.window.wait` - `Window` should not be responsible for awaiting messages, so all those usages will have to be replaced. We can re-use `waitForMessage` for that :)

src/app/communication/main-communication.js
donce: `Set` does not imply that it is set only once - i.e. setters pattern is used to set value from 0 to infinity times, not only once.
`Reload` on the other hand implies, that something identity is being loaded from somewhere - we don't need information about `identity` storage implementation.

src/app/communication/main-communication.js
donce: This is different event you're thinking about - `rendererLoaded` means that renderer process has initialized and App.vue was mounted, but it doesn't mean window is visible.

src/app/communication/main-communication.js
donce: `onCurrentIdentityChange`

donce: Renamed.

src/app/communication/utils.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+function waitForMessage (subscriber: (() => void) => void): Promise<void> {
```
shroomist: looks like a function promisification pattern.
https://hackernoon.com/transforming-callbacks-into-promises-and-back-again-e274c7cf7293

src/app/window.js
```diff
@@ -77,24 +71,18 @@ class Window {
...
-        }, timeoutInterval)
-      }
-
-      ipcMain.on(event, async (event, value) => {
-        resolve(Message(event.sender, value))
-        clearInterval(timeout)
+  // TODO: remove once it's not used anymore
+  wait (event) {
+    return new Promise((resolve) => {
+      const send = this.send.bind(this)
```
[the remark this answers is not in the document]
donce: Good catch, removed.

src/app/communication/dto.js
```diff
@@ -1,11 +1,15 @@
 // @flow
 
-export type ConnectionStatusChangeData = {
+export type ConnectionStatusChangeDTO = {
```
Waldz: +100 for DTOs

src/app/communication/main-communication.js
```diff
@@ -15,19 +15,27 @@ class MainCommunication {
...
   }
 
-  onConnectionStatusChange (callback: (ConnectionStatusChangeData) => void) {
+  onConnectionStatusChange (callback: (ConnectionStatusChangeDTO) => void): void {
     this._on(messages.CONNECTION_STATUS_CHANGED, callback)
   }
 
-  _send (channel: string, data: mixed) {
-    this._messageBus.send(channel, data)
+  onCurrentIdentityChange (callback: (IdentitySetDTO) => void) {
```
Waldz: `IdentitySetDTO ` -> `CurrentIdentityDTO`

PR #129, opened 2018-04-18
removed Mysterion factory

ignasbernotas:
Removed factory function to reduce number of imports in Mysterion class and since it wasn't needed


src/main/index.js
```diff
@@ -12,7 +18,16 @@ global.__buildNumber = process.env.BUILD_NUMBER
 global.__sentryURL = SENTRY_CONFIG.publicURL
 global.__static = MysterionConfig.staticDirectoryPath
 
-const mysterion = MysterionFactory(MysterionConfig)
+const tequilApi = new TequilAPI()
+
+const mysterion = new Mysterion({
+  config: MysterionConfig,
```
Waldz: it's `mysterionConfig `

PR #130, opened 2018-04-23
MYST-508 / New TequilApi with flow

ignasbernotas:


src/libraries/api/client/adapters/fake-adapter.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+
+import {HttpInterface} from './interface'
+
+class FakeAdapter implements HttpInterface {
```
shroomist: I think this file belongs to tests folder

src/libraries/api/client/tequil-api.js
```diff
@@ -0,0 +1,40 @@
...
+      response = await this.http.get('proposals', options)
+      if (typeof response.proposals !== 'undefined') {
+        proposals = response.proposals.map((proposal) => {
+          proposal = new ProposalDto(proposal)
+
+          return proposal
+        })
+      }
+    } catch (e) {
+      // $FlowFixMe
```
shroomist: Is it better to use `$FlowFixMe` or `TODO:` that we're already using in some places?

src/main/index.js
```diff
@@ -18,7 +18,7 @@ global.__buildNumber = process.env.BUILD_NUMBER
 global.__sentryURL = SENTRY_CONFIG.publicURL
 global.__static = MysterionConfig.staticDirectoryPath
 
-const tequilApi = new TequilAPI()
+const tequilApi = tequilApiFactory('http://127.0.0.1:4050/')
```
shroomist: shall we put the address in some config file? perhaps `MysterionConfig`?

src/libraries/api/client/dto/proposal.js
```diff
@@ -0,0 +1,32 @@
+// @flow
+
+import getCountryName from '../../../../renderer/plugins/countries'
```
shroomist: library dependency on renderer implementation
I think you can move `getCountryName` into this module

src/libraries/api/client/factory.js
```diff
@@ -0,0 +1,13 @@
+// @flow
```
shroomist: I suggest we name this file `tequilapi`, `factory` sounds too generic

src/libraries/api/client/tequil-api.js
```diff
@@ -0,0 +1,40 @@
+// @flow
+
+import {HttpInterface} from './adapters/interface'
+import ProposalDto from './dto/proposal'
+
+class TequilApi {
```
shroomist: I suggest we name this file `rest-api` or `restful`, and the factory file could be the `tequil-api` because that one could actually serve tequila.

src/libraries/api/client/factory.js
ignasbernotas: I disagree, look at the folder structure and files within. I like that the factory is separated and named accordingly:)

src/libraries/api/client/tequil-api.js
ignasbernotas: We should name files to match what they export, this class is the tequil-api client, therefore I think it suits the naming well.

src/libraries/api/client/tequil-api.js
ignasbernotas: In this case I wasn't able to fix an issue with flow due to the generics of the http adapter interface (can't find the flow's github issue to refer to), but even though the definitions are correct, flow is showing it as an error. As far as I can tell it's an issue with flow and not with the code, so I'd suggest to keep `$FlowFixMe` :)

src/libraries/api/client/adapters/axios-adapter.js
```diff
@@ -0,0 +1,38 @@
+// @flow
+
+import Axios from 'axios'
+import {HttpInterface} from './interface'
+
+class AxiosAdapter implements HttpInterface {
+  axios: Axios
```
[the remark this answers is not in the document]
Waldz: Fixed

src/libraries/api/client/dto/proposal.js
```diff
@@ -0,0 +1,24 @@
...
+    locationOriginate: {
+      country: ?string
+    }
+  }
+
+  constructor (data: Object) {
+    this.id = data.id
+    this.providerId = data.providerId
+    this.serviceType = data.serviceType
+    if (data.serviceDefinition) {
```
ignasbernotas: Not sure about this, since a proposal cannot be without a definition. Maybe throw an error if it's unset?

src/renderer/components/CountrySelect.vue
```diff
@@ -103,3 +97,6 @@
     }
   }
 </script>
+<style lang="less">
```
shroomist: empty block, delete it

test/helpers/utils.js
```diff
@@ -96,13 +6,23 @@ function nextTick () {
...
-    await func()
+    await promise
   } catch (e) {
     return e
   }
   return null
 }
 
-export default { fakeTequilapiManipulator, nextTick, captureAsyncError }
+async function captureAsyncError (func) {
```
shroomist: please add type annotations for `captureAsyncError`

shroomist: doesn't seem to be used anywhere, delete it maybe?

src/renderer/components/CountrySelect.vue
Waldz: Fixed

test/helpers/utils.js
Waldz: Instead of docblocks, rewrited test utils to Flow typed

Waldz:
All comments fixed

src/libraries/api/client/adapters/fake-adapter.js
Waldz: Fixed

src/libraries/api/client/tequil-api.js
Waldz: Improved

src/libraries/api/client/tequil-api.js
```diff
@@ -37,4 +42,12 @@ class TequilApi {
   }
 }
 
+class ProposalsFilter {
```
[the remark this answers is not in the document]
Waldz: Improved

src/libraries/api/client/dto/proposal.js
Waldz: Moved to `countries` package

src/main/index.js
Waldz: Currently moved to DI container. Possible to create config engine, then DI is good enought

src/renderer/plugins/countries.js
```diff
@@ -240,14 +242,19 @@ const countries = {
...
 
-function GetName (code) {
-  if (typeof countries[code] !== 'undefined') {
-    return countries[code]
+/**
+ * @param countryCode string
+ *
+ * @returns string
+ */
+function getCountryName (countryCode: string) {
```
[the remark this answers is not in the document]
Waldz: Implemented

ignasbernotas:
LGTM

src/libraries/api/client/dto/proposal.js
donce: If this DTO can contain `undefined` values, when we have to be aware about that everywhere we're using it - which is bad.
I agree that constructor should not throw errors.

What about having an `isValid(): boolean` method in `ProposalDTO`, which would return `true` if all required values are in place? We could check this value right after constructing dto:

```js
const dto = new ProposalDTO(...)
if (!dto.isValid()) {
  throw new Error("Missing required proposal fields")
}
```

src/libraries/api/client/adapters/interface.js
```diff
@@ -0,0 +1,17 @@
...
+}
+
+interface HttpInterface {
+  get(path: string, query: ?HttpQueryParams, options: ?Object): Promise<?mixed>;
+
+  post(path: string, data: mixed, options: ?Object): Promise<?mixed>;
+
+  delete(path: string, options: ?Object): Promise<?mixed>;
+
+  put(path: string, options: ?Object): Promise<?mixed>;
```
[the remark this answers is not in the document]
Waldz: Fixed

test/unit/app/countries/index.spec.js
```diff
@@ -0,0 +1,69 @@
+import {getCountryName, getCountryNameFromProposal, getCountryCodeFromProposal} from '../../../../src/app/countries'
+
+describe('Countries', () => {
+  describe('getCountryName()', () => {
+    it('existing country', () => {
```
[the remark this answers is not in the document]
Waldz: Fixed

src/app/countries/index.js
```diff
@@ -0,0 +1,39 @@
...
+    return null
+  }
+  if (typeof proposal.serviceDefinition.locationOriginate.country === 'undefined') {
+    return null
+  }
+
+  return proposal.serviceDefinition.locationOriginate.country
+}
+
+export {getCountryName, getCountryNameFromProposal, getCountryCodeFromProposal}
```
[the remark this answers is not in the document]
Waldz: It's public/tested function in package, can be used any-there in project

src/libraries/api/client/dto/proposals-response.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+import ProposalDTO from './proposal'
+
+type ResponseMap = {
```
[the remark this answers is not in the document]
Waldz: Renamed to `ResponseMap`

test/unit/libraries/api/client/dto/proposal-response.spec.js
```diff
@@ -0,0 +1,27 @@
...
+      expect(response.proposals[0].id).to.equal(100)
+    })
+
+    it('sets empty properties', async () => {
+      const response = new ProposalsResponseDTO({})
+
+      expect(response.proposals).to.be.undefined
+    })
+
+    it('sets wrong properties', async () => {
```
donce: `it('sets empty propoerties with wrong params'` ?

Waldz: Improved

src/libraries/api/client/dto/proposal.js
Waldz: Problem is not related to `if (data.serviceDefinition)`, undefined can happen for all fields e.g. `this.id = data.id = undefined`

PR #132, opened 2018-04-23
feature/MYST-376: display error strip when proposal list is empty

shroomist:
and if proposal fetching throws


src/renderer/components/CountrySelect.vue
```diff
@@ -83,9 +83,9 @@
       },
       async fetchCountries () {
         this.countriesAreLoading = true
-
         try {
           const response = await tequilapi.proposal.list()
+          if (response.proposals.length < 1) throw new Error('Empty proposal list received')
```
Waldz: Be consistent then formatting code

test/unit/specs/components/CountrySelect.spec.js
```diff
@@ -46,39 +49,91 @@ const tequilapiMock = () => {
...
+        [type.SHOW_ERROR] (state, error) {
+          state.errorMessage = error.message
+        }
+      }
+    }
+    beforeEach(() => {
+      store = new Vuex.Store(storeObj)
+    })
+
+    it('commits \'' + messages.countriesLoadingFailed + '\' when empty proposal list is received', async () => {
```
Waldz: Strange escaping

ignasbernotas: You can use string literals here `commits ${messages.countriesLoadingFailed} when empty proposal list is received`

src/renderer/components/CountrySelect.vue
Waldz: Empty list is not error.
And if we need this logic so much it should be in on Mysterion side

test/unit/specs/components/CountrySelect.spec.js
```diff
@@ -46,39 +49,91 @@ const tequilapiMock = () => {
...
+      expect(wrapper.vm.$store.state.errorMessage).to.eql(messages.countriesLoadingFailed)
+    })
 
-    expect(wrapper.vm.countriesList).to.have.lengthOf(4)
-    expect(wrapper.findAll('.multiselect__option-title')).to.have.lengthOf(4)
-    expect(wrapper.text()).to.contain('Lithuania')
-    expect(wrapper.text()).to.contain('United Kingdom')
-    expect(wrapper.text()).to.contain('0x3')
-    expect(wrapper.text()).to.contain('N/A')
+    it('commits \'' + messages.countriesLoadingFailed + '\' when proposal listing throws', async () => {
```
ignasbernotas: string literals here as well :)

test/unit/specs/components/CountrySelect.spec.js
```diff
@@ -46,39 +49,91 @@ const tequilapiMock = () => {
...
+      CountrySelect.__Rewire__('tequilAPI', tequilapiMockThrows)
```
Waldz: Would be good to rebase from `feature/MYST-376-update-proposals`

src/renderer/components/CountrySelect.vue
shroomist: not throwing Error anymore, but displaying Error strip with text
`'No countries available'`

PR #136, opened 2018-04-27
fix data passing via onFirstEvent from messageBus

shroomist:
mysterion.acceptTerms

### DIFF
--- a/src/app/mysterion.js
+++ b/src/app/mysterion.js
@@ -131,7 +131,7 @@ class Mysterion {
     })
 
     const termsAnswer = await this.window.wait(communication.TERMS_ANSWERED)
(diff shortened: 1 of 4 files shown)


src/app/mysterion.js
```diff
@@ -131,7 +131,7 @@ class Mysterion {
     })
 
     const termsAnswer = await this.window.wait(communication.TERMS_ANSWERED)
```
Waldz: Strange that function is limited only to return 1st parameter of event.
I would change signature to (with following request)
```
onFirstEvent(communication.TERMS_ANSWERED, () => (termsAnswer) {
  if (!termsAnswer) {
    doJobAfterTermsAccepted()
  }
})
```

PR #139, opened 2018-05-04
fix linter in tests

shroomist:

### DIFF
--- a/test/.eslintrc
+++ b/test/.eslintrc
@@ -8,5 +8,8 @@
     "should": true,
     "step": true,
     "__static": true
+  },
+  "rules": {
+    "no-unused-expressions": 0
   }
 }
--- a/test/unit/app/data-fetchers/proposal-fetcher.spec.js
+++ b/test/unit/app/data-fetchers/proposal-fetcher.spec.js
@@ -1,4 +1,4 @@
-/* eslint no-unused-expressions: 0 */
+// @flow
 import {describe, it, expect} from '../../../helpers/dependencies'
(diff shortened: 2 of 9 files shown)


test/.eslintrc
```diff
@@ -8,5 +8,8 @@
     "should": true,
     "step": true,
     "__static": true
+  },
+  "rules": {
+    "no-unused-expressions": 0
```
Waldz: Why are we disabling, it's important check?

ignasbernotas: we shouldn't disable it. maybe it's possible to use .eslintignore or similar to disable rules for certain files.

Waldz: How to solve this problem correctly?
```
http://eslint.org/docs/rules/no-unused-expressions  Expected an assignment or function call and instead saw an expression
  test/unit/libraries/api/client/dto/service-definition.spec.js:24:7
        expect(service.locationOriginate).to.be.undefined
```

shroomist: thanks for taking your time to explain this.
the problem is that I see 2 different options to toggle here:
1. specific rules of eslint for subdirectories, defilned in `.eslintrc`
2. switching off eslint for specific files in `.eslintignore`
and not the combination of the two.
I can put `"no-unused-expressions": 0` in `unit` and `integration` folders, that should also serve fine, that should keep files in `helpers` folder strict, but `karma.conf` files would still suffer of being sloppy, we might as well switch this rule :on in header of `karma.conf` files, I think.

PR #143, opened 2018-05-07
MYST-508 Replace tequilapi connection endpoints

Waldz:


src/renderer/store/modules/connection.js
```diff
@@ -1,76 +1,122 @@
...
+  action: string
+  looper: FunctionLooper
+
+  constructor (action: string, looper: FunctionLooper) {
+    this.action = action
+    this.looper = looper
+  }
+}
+
+class ActionLooperStart {
```
shroomist: perhaps `ActionLooperSettings` or `ActionLooperConfig` would be more clear as a class name

Waldz: Renamed

PR #146, opened 2018-05-08
MYST-508 Separate Mysterion and MysteriumClient configs

Waldz:


src/main/dependencies/modules/mysterium-client.js
```diff
@@ -0,0 +1,46 @@
...
+    ['mysterionApplication.config'],
+    (mysterionConfig: MysterionConfig) => {
+      return new ClientConfig(
+        path.join(mysterionConfig.contentsDirectory, 'bin', 'mysterium_client'),
+        path.join(mysterionConfig.contentsDirectory, 'bin', 'config'),
+        path.join(mysterionConfig.contentsDirectory, 'bin', 'openvpn'),
+        mysterionConfig.userDataDirectory,
+        mysterionConfig.runtimeDirectory,
+        mysterionConfig.userDataDirectory
+      )
```
donce: Long list of parameters is hard to understand and maintain, and it's easy to add a bug in it.
Maybe flow `type Config` would work better here instead of `class Config`?
Then construction would be more understandable:
```js
{
  clientBin: path.join(mysterionConfig.contentsDirectory, 'bin', 'mysterium_client'),
  configDir: path.join(mysterionConfig.contentsDirectory, 'bin', 'config')
  ...
```

src/main/dependencies/modules/mysterium-client.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+  )
+  container.service(
+    'mysteriumClientInstaller',
+    ['mysteriumClient.config'],
+    (config) => {
+      return new Installer(config)
+    }
```
donce: We can use shorter form here:
```js
(config) => new Installer(config)
```

src/main/dependencies/modules/mysterium-client.js
```diff
@@ -0,0 +1,46 @@
...
+import type {MysterionConfig} from '../../../app/mysterionConfig'
+import {Installer, Process, Monitoring} from '../../../libraries/mysterium-client'
+import path from 'path'
+
+function bootstrap (container: Container) {
+  container.service(
+    'mysteriumClient.config',
+    ['mysterionApplication.config'],
+    (mysterionConfig: MysterionConfig) => {
+      return new ClientConfig(
```
donce: We should use `type` instead of `class` for `ClientConfig` - then we'll avoid long parameters list, which is dangerous. Than we could initiate ClientConfig using:
```js
const config: ClientConfig = {
  clientBin: path.join(mysterionConfig.contentsDirectory, 'bin', 'mysterium_client'),
  configDir: path.join(mysterionConfig.contentsDirectory, 'bin', 'config')
  ...
```

Waldz: Constructor makes sense here, it describes required parameters.
Also defaults some parameters to default values

Waldz: Fixed

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -30,14 +23,31 @@ class Process {
...
   }
+
+  /**
+   * Converts log level to log files 'stdout.log' or 'stderr.log'
+   *
+   * @param {string} level
+   * @return {string}
+   * @private
+   */
+  _getFileForLevel (level) {
```
donce: Nice abstraction 👍

src/libraries/mysterium-client/standalone/process.js
```diff
@@ -29,17 +28,33 @@ class Process {
...
   }
+
+  /**
+   * Converts log level to child process members 'child.stdout' and 'child.stderr'
+   *
+   * @param {string} level
+   * @return {Readable}
+   * @private
+   */
+  _getStreamForLevel (level) {
```
donce: When we have this file in flow, we will be able to define `level` param to be enum. In that case we'll probably won't have to do `default` case of `switch` :)

src/main/dependencies/modules/mysterion.js
```diff
@@ -0,0 +1,108 @@
...
+        process: mysteriumClientProcess,
+        monitoring: mysteriumClientMonitoring,
+        proposalFetcher: proposalFetcher,
+        bugReporter: bugReporter
+      })
+    }
+  )
+
+  const browserWindowIsSingleton = true
+  let browserWindow
```
donce: This doesn't have to be scoped outside of `.factory` - it's not used anywhere. Also, same variable name is used as a parameter in line 98, which is very confusion and is a code smell.

src/main/dependencies/modules/mysterion.js
```diff
@@ -0,0 +1,108 @@
...
+      })
+      return browserWindow
+    },
+    browserWindowIsSingleton
+  )
+
+  container.service(
+    'mysterionWindow',
+    ['mysterionBrowserWindow', 'feedbackForm.headerRule'],
+    (browserWindow, rule) => {
```
donce: `browserWindow` parameter overrides `let browserWindow` in line 81.

PR #149, opened 2018-05-09
Connecting via tray updates dropdown selected country

ignasbernotas:


src/app/countries/index.js
```diff
@@ -36,27 +36,42 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
...
 
-    return country
-  })
+  country.label = country.label + ' (' + idLabel + ')'
+
+  return country
+}
+
+function getCountryListFromProposals (proposals: Array<ProposalDTO>): Array<Country> {
+  let countries = proposals.map((proposal) => getCountryFromProposal(proposal))
```
shroomist: this does append iD to all of the entries in the country list, doesn't it?
we need it for duplicated only..

src/renderer/components/CountrySelect.vue
```diff
@@ -55,6 +55,11 @@
       },
       countryLabel (country) {
         if (typeof country === 'object') {
+          const allowedLength = 22
+          if (country.label.length > allowedLength) {
```
shroomist: you're gonna be cutting off the `...` part here for longer named countries.

shroomist: you could test it with country `The Democratic Republic of the Congo`

larashed: This for the selected item only. Without the cut it'll skew the design.

src/app/countries/index.js
larashed: Not sure what you mean. We need the ID to be able to connect to each node.

ignasbernotas: Not sure what you mean. We need the ID to be able to connect to each node.

src/renderer/components/CountrySelect.vue
ignasbernotas: This is for the selected item only. If the label length isn't reduced it skews the design.

src/app/countries/index.js
shroomist: looks like you are adding `idLabel` to each country label, but requirement is to do it for duplicates only

ignasbernotas: Not anymore, we decided not to group the countries due to large scope of the task.

src/renderer/components/CountrySelect.vue
```diff
@@ -84,6 +89,17 @@
 
         this.countriesList = getCountryListFromProposals(proposals)
       })
+
+      this.rendererCommunication.onConnectionRequest((proposal) => {
```
Waldz: Is it proposal in `.onConnectionRequest((proposal)` or country?

src/app/countries/index.js
```diff
@@ -36,27 +36,42 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
...
+    label: getCountryNameFromProposal(proposal),
+    id: proposal.providerId,
+    code: getCountryCodeFromProposal(proposal)
+  }
 
-    let idLabel = country.id
-    if (country.id.length > idLength) {
-      idLabel = country.id.substring(0, 10) + '...'
-    }
+  const idLength = 10
```
Waldz: Move it as constant in package

src/renderer/components/CountrySelect.vue
```diff
@@ -84,6 +89,17 @@
...
+        let selectedCountry = this.countriesList.find((country) => country.id === proposal.providerId)
+
+        if (!selectedCountry) {
```
Waldz: Should not You show error to user on this case? Currently silently skipping user's desire

src/renderer/components/CountrySelect.vue
```diff
@@ -84,6 +89,17 @@
...
+          return
+        }
+
+        this.country = selectedCountry
```
Waldz: You should assign country in beginning on user's desire. There is a bug then user selects country from Tray

src/renderer/components/CountrySelect.vue
ignasbernotas: proposal

src/app/countries/index.js
ignasbernotas: But it's not a package constant. It's a small function scope var that won't be used elsewhere.

src/renderer/components/CountrySelect.vue
ignasbernotas: @Waldz Not sure how to handle it though. User selects tray, but if country no longer exists in the dropdown there's not much we can do. If the country is no longer available, user will not get connected anyway and will be shown an error, so adding another error isn't exactly an option.

ignasbernotas: Fixed. If the country isn't found after clicking tray, the dropdown is reset into "Choose country"

donce:
Please review comments from https://github.com/MysteriumNetwork/mysterion/pull/147

src/app/countries/index.js
```diff
@@ -36,4 +36,53 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
   return proposal.serviceDefinition.locationOriginate.country
 }
 
-export {getCountryName, getCountryNameFromProposal, getCountryCodeFromProposal}
+type Country = {
+  label: string,
+  id: string,
+  code: string
```
donce: Is `code` actually used anywhere?

src/app/countries/index.js
```diff
@@ -36,4 +36,53 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
...
+  if (a.label > b.label) {
+    comparison = 1
+  } else if (b.label > a.label) {
+    comparison = -1
+  }
+
+  return comparison
+}
+
+function getCountryListFromProposals (proposals: Array<ProposalDTO>): Array<Country> {
```
donce: `getSortedCountriesFromProposals` ?

src/app/countries/index.js
```diff
@@ -36,4 +36,53 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
...
+  code: string
+}
+
+function getCountryFromProposal (proposal: ProposalDTO): Country {
+  const allowedIdentityLength = 10
+
+  let country = {
+    id: proposal.providerId,
+    code: getCountryCodeFromProposal(proposal)
+  }
```
donce: Instead of building country with missing `label` field and mutating it later, we can:
```js
const label = +  country.label = getCountryNameFromProposal(proposal) + ' (' + identity + ')'

let country = {
  id: proposal.providerId,
  code: getCountryCodeFromProposal(proposal),
  label
}
```

src/app/countries/index.js
```diff
@@ -36,4 +36,53 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
...
+  if (identity.length > allowedIdentityLength) {
+    identity = identity.substring(0, allowedIdentityLength) + '...'
+  }
+
+  country.label = getCountryNameFromProposal(proposal) + ' (' + identity + ')'
+
+  return country
+}
+
+function _sortCountries (a, b) {
```
donce: Method name is lying about function - it does not do any sorting.
`_compareCountries`?

donce: Can we specify types for countries here, that is `a: Country, b: Country`?

src/app/countries/index.js
```diff
@@ -36,4 +36,53 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
...
+    id: proposal.providerId,
+    code: getCountryCodeFromProposal(proposal)
+  }
+
+  let identity = proposal.providerId
+  if (identity.length > allowedIdentityLength) {
+    identity = identity.substring(0, allowedIdentityLength) + '...'
+  }
+
+  country.label = getCountryNameFromProposal(proposal) + ' (' + identity + ')'
```
donce: String interpolation could be used here:
```js
const countryName = getCountryNameFromProposal(proposal)
country.label = `${countryName} (${identity})`
```

src/app/countries/index.js
```diff
@@ -36,4 +36,53 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
...
+
+  return country
+}
+
+function _sortCountries (a, b) {
+  let comparison = 0
```
donce: Why do we need this variable? Can't we just return the value?

```
if (a.label > b.label) {
  return 1
} else if (b.label > a.label) {
  return -1
}
return 0
```

src/app/countries/index.js
```diff
@@ -36,4 +36,53 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
...
+    comparison = 1
+  } else if (b.label > a.label) {
+    comparison = -1
+  }
+
+  return comparison
+}
+
+function getCountryListFromProposals (proposals: Array<ProposalDTO>): Array<Country> {
+  const countries = proposals.map((proposal) => getCountryFromProposal(proposal))
```
donce: We don't need to construct lambda function, we can pass `getCountryFromProposal` directly:
```js
const countries = proposals.map(getCountryFromProposal)
```

src/renderer/components/CountrySelect.vue
donce: Maybe we should reduce country name to solve this issue?
```
The Demog... (0x1251f25e)
```
Doesn't seem nice, but at least we're consistent with providing identity.

src/renderer/components/CountrySelect.vue
```diff
@@ -90,7 +87,14 @@
           return
         }
 
-        this.countriesList = proposals.map(proposalToCountry)
+        this.countriesList = getCountryListFromProposals(proposals)
```
donce: `countriesList` is not english - it should be `countryList` or `countries`.

src/app/countries/index.js
ignasbernotas: In CountrySelect.vue

src/renderer/components/CountrySelect.vue
ignasbernotas: Actually it is:)

donce: AFAIK, it's translation from lithuanian (*šalių sąrašas*), but it's uncommon to use plural in english.

Didn't find any good reference shortly, but here are the same question here: https://www.proz.com/forum/linguistics/245800-noun_list_is_noun_singular_or_plural.html

src/app/countries/index.js
ignasbernotas: fixed in a different way

src/app/countries/index.js
ignasbernotas: fixed

src/app/countries/index.js
ignasbernotas: fixed

src/app/countries/index.js
ignasbernotas: fixed

src/renderer/components/CountrySelect.vue
```diff
@@ -53,18 +45,30 @@
...
+        if (typeof country !== 'object') {
+          return
         }
+
+        const allowedLength = 22
+
+        if (country.label.length > allowedLength) {
+          let countryName = country.name
+          if (countryName.length > 12) {
+            countryName = `${countryName.substring(0, 10)}..`
```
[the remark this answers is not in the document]
ignasbernotas: fixed

src/app/countries/index.js
```diff
@@ -36,4 +36,53 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
...
+  let identity = proposal.providerId
+  if (identity.length > allowedIdentityLength) {
+    identity = identity.substring(0, allowedIdentityLength) + '..'
+  }
+
+  const countryName = getCountryNameFromProposal(proposal)
+  let country = {
+    id: proposal.providerId,
+    code: getCountryCodeFromProposal(proposal),
+    label: `${countryName} (${identity})`,
```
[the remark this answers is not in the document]
ignasbernotas: moved it out of struct into a separate getCountryLabel function, but still in app/countries/index.js

src/renderer/components/CountrySelect.vue
```diff
@@ -90,7 +94,14 @@
           return
         }
 
-        this.countriesList = proposals.map(proposalToCountry)
+        this.countryList = getSortedCountryListFromProposals(proposals)
+      })
+
+      this.rendererCommunication.onConnectionRequest((proposal) => {
+        let selectedCountry = this.countryList.find((country) => country.id === proposal.providerId)
```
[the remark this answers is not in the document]
ignasbernotas: fixed

src/app/countries/index.js
```diff
@@ -36,4 +36,53 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
...
+
+  let identity = proposal.providerId
+  if (identity.length > allowedIdentityLength) {
+    identity = identity.substring(0, allowedIdentityLength) + '..'
+  }
+
+  const countryName = getCountryNameFromProposal(proposal)
+  let country = {
+    id: proposal.providerId,
+    code: getCountryCodeFromProposal(proposal),
```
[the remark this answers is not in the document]
ignasbernotas: fixed

test/unit/renderer/components/CountrySelect.spec.js
```diff
@@ -128,4 +129,35 @@ describe('CountrySelect', () => {
...
+
+    it('returns truncated long country name label', () => {
+      const country = {
+        label: 'The Democratic Republic of the Congo (0x1234567890)',
+        name: 'The Democratic Republic of the Congo',
+        id: '0x1234567890',
+        code: 'cd'
+      }
+
+      let label = wrapper.vm.selectedCountryLabel(country)
```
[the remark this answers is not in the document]
ignasbernotas: fixed

test/unit/renderer/components/CountrySelect.spec.js
```diff
@@ -128,4 +129,35 @@ describe('CountrySelect', () => {
...
+
+    it('returns truncated short country name label', () => {
+      const country = {
+        label: 'Lithuania (0x1234567890)',
+        name: 'Lithuania',
+        id: '0x1234567890',
+        code: 'lt'
+      }
+
+      let label = wrapper.vm.selectedCountryLabel(country)
```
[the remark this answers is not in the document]
ignasbernotas: fixed

ignasbernotas:
@donce @Waldz @shroomist please review

src/app/countries/index.js
```diff
@@ -36,4 +36,52 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
...
+function getCountryLabel (country: Country, maxIdentityLength: ?number = 9, maxNameLength: ?number = null) {
+  let identity = country.id
+  if (maxIdentityLength && identity.length > maxIdentityLength) {
+    identity = identity.substring(0, maxIdentityLength) + '..'
+  }
+
+  let name = country.name
+  if (maxNameLength && name.length > maxNameLength) {
+    name = name.substring(0, maxNameLength) + '..'
+  }
```
donce: This part is duplicated, it could be DRYied up a bit:
```js

function getCountryLabel (country: Country, maxIdentityLength: ?number = 9, maxNameLength: ?number = null) {
  const identity = _limitedLengthString(country.id, maxIdentityLength)
  const name = _limitedLengthString(country.name, maxNameLength)

  return `${name} (${identity})`
}

function _limitedLengthString(value: string, maxLength: ?number = null) {
  if (maxLength && value.length > maxLength) {
    return value.substring(0, maxLength) + '..'
  }
  return value
}
```

src/renderer/components/CountrySelect.vue
```diff
@@ -53,18 +45,27 @@
...
     methods: {
       onChange (country) {
         this.$emit('selected', country)
       },
+      selectedCountryLabel (country) {
+        if (typeof country !== 'object') {
+          return
+        }
+
+        return getCountryLabel(country, 9, 10)
```
donce: If we switch 2nd and 3rd arguments, we wouldn't have to duplicate `9` constant in two separate files.
```js
getCountryLabel(country: Country, maxNameLength: ?number = null, maxIdentityLength: ?number = 9)
getCountryLabel(country, 10)
```

src/app/countries/index.js
```diff
@@ -4,7 +4,7 @@ import ProposalDTO from '../../libraries/mysterium-tequilapi/dto/proposal'
 
 const COUNTRY_NAME_UNKNOWN = 'N/A'
 
-function getCountryName (countryCode: string): string {
+function _getCountryName (countryCode: string): string {
```
shroomist: I don't think you need to prepend functions with `_` here.
this practice is used to hide class members, mark them as private.
in this case functions stay in this module scope, and its has clear export definition in the bottom.

src/app/countries/index.js
ignasbernotas: fixed

src/renderer/components/CountrySelect.vue
ignasbernotas: fixed

src/app/countries/index.js
ignasbernotas: fixed

test/unit/renderer/components/CountrySelect.spec.js
```diff
@@ -128,4 +128,33 @@ describe('CountrySelect', () => {
...
+
+    it('returns truncated long country name label', () => {
+      const country = {
+        name: 'The Democratic Republic of the Congo',
+        id: '0x1234567890',
+        code: 'cd'
+      }
+
+      const label = wrapper.vm.selectedCountryLabel(country)
+      expect(label).to.be.eql('The Democr.. (0x1234567..)')
```
shroomist: so basically we leave only 10 chars for country name..
`United Kin.. (0x1234567..)`
I suggest we don't add `..` after `id` and drop `0x` part.
we can win 3-4 chars this way.
`United Kingdom (1234567)`

ignasbernotas: your example works for a single country, other countries will still have the same issue

PR #150, opened 2018-05-11
Refactor IPC communication

donce:


src/renderer/pages/Terms.vue
```diff
@@ -14,19 +14,23 @@
 </template>
 <script>
   import {mapGetters} from 'vuex'
-  import communication from '../../app/communication'
   import RendererMessageBus from '../../app/communication/rendererMessageBus'
+  import RendererCommunication from '../../app/communication/renderer-communication'
+
+  function answerTerms (answer) {
+    const messageBus = new RendererMessageBus()
```
Waldz: Don't construct in the middle of code, use DI to get same instance

donce: Fixed.

PR #153, opened 2018-05-15
Re-enable flow for .js files

donce:
Flow checks for *.js* files were accidentally disable - it was working only for *.vue*.
This PR:
- Re-enables Flow checks for *.js* files
- Makes Flow aware of '@' alias in import statements
- Fixes some of Flow warnings
- Temporary removes Flow checking from CI - until we clean-up all errors. There are many errors, so fixing all of them with one PR is an overkill and will cause too many conflicts.

### DIFF
--- a/src/renderer/store/modules/connection.js
+++ b/src/renderer/store/modules/connection.js
@@ -19,7 +19,7 @@ type ConnectionStore = {
   ip: ?string,
   status: ConnectionStatus,
   statistics: Object,
-  actionLoopers: Map<string, FunctionLooper>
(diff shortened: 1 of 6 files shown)


src/renderer/store/modules/connection.js
```diff
@@ -181,7 +181,7 @@ function actionsFactory (
         }
 
         commit(type.SHOW_ERROR_MESSAGE, messages.connectFailed)
-        let error = new Error('Connection to node failed.')
+        const error: Object = new Error('Connection to node failed.')
```
shroomist: maybe Error type here?

PR #156, opened 2018-05-15
Do not release same version twice when tagging semantic version

donce:
From https://docs.travis-ci.com/user/customizing-the-build:

> Note that `before_deploy` and `after_deploy` are run before and after every deploy provider, so will run multiple times if there are multiple providers.


donce:
@ignasbernotas AFAIK, we should be able to remove git tag and add it on a new commit to update release of existing tag if that's really what we want - this PR shouldn't change this behavior.

PR #159, opened 2018-05-16
Bugfix/mys 44 tequilapi client error

shroomist:
due to transpiling issues, and the fact that we're using this both on front and back end we cannot simply extend native Error type.


src/libraries/mysterium-tequilapi/client-error.js
```diff
@@ -9,38 +9,46 @@ const httpResponseCodes = {
...
-    super.name = this.constructor.name
-    super.message = original.message
-    super.stack = original.stack
-
-    this.original = original
+function TequilapiClientError (original: Error) {
+  const instance = Reflect.construct(Error, [original.message])
+  instance.name = 'TequilapiClientError'
+  Reflect.setPrototypeOf(instance, Reflect.getPrototypeOf(this))
+  Error.captureStackTrace(instance, TequilapiClientError)
```
donce: I am very sceptical about using reflect for extending error class - I think we should avoid extending error class instead. It might not be convenient, it's hard to tell - maybe I should try to implement that?

ignasbernotas: I agree, this seems like a lot of trouble for a hack. Need to investigate how other projects achieve this.

shroomist: I'd like to see having `ourError instanceof Error === true`.
you can give it a try.

donce: Done: https://github.com/MysteriumNetwork/mysterion/pull/168

PR #160, opened 2018-05-16
di-factory instance caching is now switched on by default

shroomist:


ignasbernotas:
Please rebase before merging.

PR #161, opened 2018-05-16
MYST-508 Mysterion logs

Waldz:


src/app/mysterion.js
```diff
@@ -26,26 +28,40 @@ class Mysterion {
...
       }
-      this.window.show()
     })
     app.on('before-quit', () => {
       this.window.willQuitApp = true
     })
   }
 
   async bootstrap () {
+    logInfo('Checking terms cache')
```
ignasbernotas: Change to "Bootstrapping application"?

src/app/mysterion.js
```diff
@@ -60,60 +76,71 @@ class Mysterion {
       browserWindow = this.browserWindowFactory()
       this.window = this.windowFactory()
       if (termsAccepted) {
+        logInfo('Opening application window')
```
ignasbernotas: "Terms accepted, resizing the window to app window size"

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -17,9 +17,13 @@ class Process {
   }
 
   start () {
-    this.tequilapi.healthCheck().catch(() => {
-      console.log('touched the daemon, now it should be up')
-    })
+    this.tequilapi.healthCheck()
+      .then(() => {
+        console.log('Touched the daemon with error, anyway it should be up')
```
ignasbernotas: Shouldn't these be switched places? Catch callback should say "with error"

test/integration/libraries/mysterium-client/standalone/process.spec.js
```diff
@@ -36,22 +37,44 @@ xdescribe('Standalone Process', () => {
     process.stop()
   })
 
-  it('spawns in less than 100ms without errors', () => {
-    expect(process.child).to.be.instanceOf(ChildProcess)
-  })
+  describe('startup', () => {
+    it('spawns in less than 100ms without errors', () => {
```
ignasbernotas: not sure about these 100ms? Maybe increase it to 500 at least?

src/app/mysterion.js
```diff
@@ -26,26 +28,40 @@ class Mysterion {
...
-        return this.bootstrap()
+      try {
+        logInfo('Application activate')
+        if (!this.window.exists()) {
+          this.bootstrap()
+          return
+        }
+        this.window.show()
+      } catch (e) {
+        logException('Application activate failed', e)
```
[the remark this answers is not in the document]
Waldz: Changed

src/app/mysterion.js
Waldz: But i want to log terms step here

src/app/mysterion.js
Waldz: Changed

src/libraries/mysterium-client/launch-daemon/process.js
Waldz: Fixed

test/integration/libraries/mysterium-client/standalone/process.spec.js
Waldz: Not my scope

src/app/mysterion.js
```diff
@@ -25,26 +27,40 @@ class Mysterion {
...
-        return this.bootstrap()
+      try {
+        logInfo('Show window')
+        if (!this.window.exists()) {
+          this.bootstrap()
+          return
+        }
+        this.window.show()
+      } catch (e) {
+        logException('Failed to show window', e)
```
donce: This is false - there are more things in `try` block than showing a window, i.e. terms, process initialisation and so on.

src/app/mysterion.js
```diff
@@ -25,26 +27,40 @@ class Mysterion {
...
+        this.bugReporter.main.captureException(e)
```
donce: I'm not 100% sure, but I believe that failed hook will terminate application if this hook throws an exception.
Are you capturing exception here intentionally? Can't this leave dead app hanging in some failure cases?

src/app/mysterion.js
```diff
@@ -25,26 +27,40 @@ class Mysterion {
...
-      await this.bootstrap()
-      this.buildTray()
+      try {
+        logInfo('Application launch')
+        await this.bootstrap()
+        this.buildTray()
+      } catch (e) {
+        logException('Application launch failed', e)
+        this.bugReporter.main.captureException(e)
+      }
```
donce: I'm not 100% sure, but I believe that failed hook will terminate application if this hook throws an exception.
Are you capturing exception here intentionally? Can't this leave dead app hanging in some failure cases?

src/app/mysterion.js
```diff
@@ -59,60 +75,71 @@ class Mysterion {
...
     if (this.installer.needsInstallation()) {
+      logInfo("Installing 'mysterium_client' process")
       try {
         await this.installer.install()
       } catch (e) {
-        this.bugReporter.captureException(e)
-        console.error(e)
-        return this.communication.sendRendererShowErrorMessage(translations.daemonInstallationError)
+        this.communication.sendRendererShowErrorMessage(translations.daemonInstallationError)
+        throw new Error("Failed to install 'mysterium_client' process. " + e)
```
donce: We should not create a new instance of `Error` here - stacktrace of original error is lost here, which provides details of where error happened.

src/app/mysterion.js
```diff
@@ -153,62 +180,64 @@ class Mysterion {
...
-  /**
-   * notifies the renderer that we're good to go and sets up the system tray
-   */
-  startApp () {
+  updateProposalsLoop () {
     this.proposalFetcher.subscribe((proposals) => this.communication.sendProposals(proposals))
-    this.proposalFetcher.start()
+
+    this.monitoring.subscribeUp(() => {
+      logInfo("Subscribing 'mysterium_client' proposals")
```
donce: This is confusing - nothing is being subscribed here. `Starting proposal fetcher`?

src/app/mysterion.js
```diff
@@ -153,62 +180,64 @@ class Mysterion {
...
+    await this.process.start()
     this.process.onLog(processLogLevels.LOG, (data) => cacheLogs(processLogLevels.LOG, data))
     this.process.onLog(processLogLevels.ERROR, (data) => cacheLogs(processLogLevels.ERROR, data))
+  }
 
-    this.monitoring.start()
-    this.monitoring.onProcessReady(() => {
-      updateRendererWithHealth()
-      this.startApp()
+  startProcessMonitoringLoop () {
```
donce: why monitoring *loop*? There doesn't seem to be any loop mentioned inside - it's just an implementation detail which is hidden from us. Maybe simply `startProcessMonitoring`?

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -17,9 +17,13 @@ class Process {
   }
 
   start () {
-    this.tequilapi.healthCheck().catch(() => {
-      console.log('touched the daemon, now it should be up')
-    })
+    this.tequilapi.healthCheck()
+      .then(() => {
+        console.log('Touched the daemon, now it must be up')
```
donce: `should` sounded better - it's possible that it's not up.

src/libraries/mysterium-client/monitoring.js
```diff
@@ -1,55 +1,95 @@
...
-    if (this.ipcTimeout) {
-      clearTimeout(this.ipcTimeout)
+    if (this._timer) {
+      clearTimeout(this._timer)
     }
   }
 
-  isRunning () {
-    return this.clientIsRunning
+  subscribeStatus (callback: StatusCallback) {
```
donce: This doesn't seem to be used.

src/libraries/mysterium-client/monitoring.js
```diff
@@ -1,55 +1,95 @@
...
     try {
       await this.api.healthCheck(healthCheckTimeout)
-      this.clientIsRunning = true
+      isRunning = true
     } catch (e) {
-      this.clientIsRunning = false
+      isRunning = false
+    }
+
+    this._notifySubscribers(isRunning)
```
donce: If any of subscribers throws an error, healthcheck loop will stop.
We keep our healthcheck running even if one of subscribers callbacks failed:
```js
let err = null
try {
  this._notifySubscribers(isRunning)
} catch (e) {
  err = e
}
this._timer = setTimeout(() => this._healthCheckLoop(), healthCheckInterval)
if (err) {
  throw error
}
```

src/main/dependencies/modules/mysterium-client.js
```diff
@@ -34,7 +34,9 @@ function bootstrap (container: Container) {
   container.service(
     'mysteriumClientMonitoring',
     ['tequilapiClient'],
-    (tequilapiClient) => new Monitoring(tequilapiClient)
+    (tequilapiClient) => {
+      return new Monitoring(tequilapiClient)
```
donce: Why long form?

test/integration/libraries/mysterium-client/standalone/process.spec.js
```diff
@@ -36,22 +37,44 @@ xdescribe('Standalone Process', () => {
...
+    it('sends log data to callback on()', () => {
+      expect(logs.pop()).to.include('Api started on: ' + tequilapiPort)
+    })
   })
 
-  it('responds to healthcheck with uptime', async () => {
-    const res = await tequilapi.healthCheck()
-    expect(res.uptime).to.be.ok
-    expect(res.version).to.be.ok
+  describe('health checking', () => {
```
donce: Are you aware that all these tests are disabled, because of `xdescribe` in this file? Have you tried running them when they are enabled?

src/app/mysterion.js
Waldz: Fixed

src/app/mysterion.js
Waldz: Code is tested.
Exception is captured intentionally, so Sentry knows about all fatal failures

src/app/mysterion.js
Waldz: same here

src/app/mysterion.js
Waldz: There is TODO for that, not in PR scope

src/app/mysterion.js
Waldz: Improved

src/app/mysterion.js
Waldz: renamed

src/app/mysterion.js
donce: > Code is tested.
You mean you've tested it manually?

> Exception is captured intentionally, so Sentry knows about all fatal failures
Yes, but it's not re-thrown anymore - that's what is concerning me.

src/libraries/mysterium-client/monitoring.js
Waldz: Fixed

src/main/dependencies/modules/mysterium-client.js
Waldz: Fixed

test/integration/libraries/mysterium-client/standalone/process.spec.js
Waldz: Off course

src/libraries/mysterium-client/launch-daemon/process.js
Waldz: Fixed

src/app/mysterion.js
Waldz: PR review is meant for code review, not assumptions.

donce: You are changing this behaviour in you PR with code changes, that's why I want to understand is it intentional or not and how does it effect start-up.

src/app/mysterion.js
```diff
@@ -53,18 +68,28 @@ class Mysterion {
...
+      await this._acceptTermsOrQuit()
     }
 
     await this._ensureDaemonInstallation()
-    await this._startProcessAndMonitoring()
+    // if all is good, let's boot up the client
+    this._startProcess()
+    // and start monitoring it
+    this._startProcessMonitoring()
+    // and handle some actions after process is up
```
donce: These new comments in this block seem redundant - code is pretty self-explanatory.
Perhaps they are result of conflict resolving?

src/app/mysterion.js
```diff
@@ -193,62 +215,62 @@ class Mysterion {
...
 
   /**
-   * notifies the renderer that we're good to go and sets up the system tray
+   * Blocking function until process is launched
    */
-  startApp () {
+  async _onProcessReady (callback) {
+    await onFirstEvent(this.monitoring.subscribeUp.bind(this.monitoring))
+
+    callback()
```
donce: > This will go to library `mysterium-client`, will be useful for other flavours

How will it be useful? My first though was for backwards-compatibility with callbacks without async/await, but this method is `async`, so we are already required lib users to be able to support async/await.

src/libraries/mysterium-client/monitoring.js
donce: This doesn't seem to be productive - it's against [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

When are we planning to extract this library?

It seems like useful method, but at the moment, this class is not a library - now it serves for us, and the simpler it is, the faster we'll move.

Once it's a separate library, then we can add additional features, which might be required by other possible users.

test/integration/libraries/mysterium-client/standalone/process.spec.js
```diff
@@ -36,22 +37,44 @@ xdescribe('Standalone Process', () => {
...
+    before(async () => {
+      processMonitoring = new Monitoring(tequilapi)
+      processMonitoring.start()
+    })
+
+    after(() => {
+      processMonitoring.stop()
+    })
+
+    it('monitoring send UP status', (done) => {
```
donce: *"it monitoring send UP status"* doesn't make sense to me - *"it sends status UP"* ?

test/integration/libraries/mysterium-client/standalone/process.spec.js
```diff
@@ -36,22 +37,44 @@ xdescribe('Standalone Process', () => {
...
+      processMonitoring.subscribeUp(() => {
+        done()
+      })
```
donce: We can pass `done` correctly, no need to create one more function abstraction:
```js
processMonitoring.subscribeUp(done)
```

src/app/mysterion.js
donce: Which TODO are you referring? I'm not talking about some unrelated existing issues - this change brakes stacktrace.

Waldz: It's old convention in this file with TODO registered. Fix fill come in upcomming separate PR

src/libraries/mysterium-client/monitoring.js
Waldz: it's  enough to understand, because code is laying in `src/libraries/..`. Library migration will come in separate PR

src/app/mysterion.js
Waldz: Thanks. Removed

src/app/mysterion.js
Waldz: Quote not related to code lines

test/integration/libraries/mysterium-client/standalone/process.spec.js
Waldz: Improved

test/integration/libraries/mysterium-client/standalone/process.spec.js
Waldz: Improved

Waldz:
I must remind that this PR implements more detailed Mysterion startup logs ONLY

src/app/mysterion.js
Waldz: Did not get about lib users

Waldz: I just want to collect actions performed after process is ready to single method.
BTW code was the same before PR 

```

```

donce: I do not get what are you telling... Lets start from beginning - why do you need `async` method with a `callback` param?

> BTW code was the same before PR

Can you direct me to it in `master`? I don't see any `async` methods with callback parameter in removed code.

src/app/mysterion.js
```diff
@@ -193,62 +212,62 @@ class Mysterion {
...
+      this.communication.sendMysteriumClientDown()
     })
+
+    logInfo("Starting 'mysterium_client' monitoring")
+    this.monitoring.start()
   }
 
   /**
-   * notifies the renderer that we're good to go and sets up the system tray
+   * Blocking function until process is launched
```
shroomist: this comment here is confusing, 
`async _onProcessReady` is non-blocking,
I think this method signature is self-explanatory.

src/app/mysterion.js
```diff
@@ -193,62 +212,62 @@ class Mysterion {
...
    */
-  startApp () {
+  async _onProcessReady (callback) {
+    await onFirstEvent(this.monitoring.subscribeUp.bind(this.monitoring))
```
shroomist: to simplify this, you could do
```
_onProcessReady (callback) {
   onFirstEvent(this.monitoring.subscribeUp.bind(this.monitoring)).then(callback)
```

src/libraries/mysterium-client/monitoring.js
shroomist: I suggest `subscribeStatus` => `onStatusChange(callback)` or `subscribeStatusChange`

src/app/mysterion.js
donce: Exactly.

src/app/mysterion.js
ignasbernotas: @Waldz _onProcessReady is only used in that single place, and `this.communication.sendMysteriumClientIsReady()` can be moved to the method `_onProcessReady` itself, no need for the callback since it's not being reused:)

What you're referring to in master is a different case, `this.monitoring.onProcessReady` was called directly, and since this method is being called multiple times in different places it makes sense for it to have a callback function as an argument (since many different places need to execute some logic during this event), but it isn't necessary to pass a callback in a small abstraction method (`_onProcessReady`) that isn't reused.

src/libraries/mysterium-client/monitoring.js
ignasbernotas: 👍 for `onStatusChange`

src/app/mysterion.js
```diff
@@ -150,27 +171,25 @@ class Mysterion {
...
   // declining terms will quit the app
   async _acceptTermsOrQuit () {
+    logInfo('Accepting terms')
     try {
       const accepted = await this._acceptTerms()
       if (!accepted) {
-        console.log('Terms were refused. Quitting.')
+        logInfo('Terms were refused. Quitting.')
         app.quit()
-        return false
```
donce: We should terminate `bootstrap` instantly if user declines terms, we shouldn't start-up mysterium client process and other things. AFAIK, `app.quit()` does not stop bootstrap.

src/app/mysterion.js
```diff
@@ -193,62 +212,62 @@ class Mysterion {
...
-    }
+  async _startProcess () {
     const cacheLogs = (level, data) => {
       this.communication.sendMysteriumClientLog({level, data})
       this.bugReporter.pushToLogCache(level, data)
     }
 
-    this.process.start()
+    logInfo("Starting 'mysterium_client' process")
+    await this.process.start()
```
donce: `start` is not async, you don't have to `await` it.

src/app/mysterion.js
donce: Still not resolved.

src/libraries/mysterium-client/monitoring.js
```diff
@@ -1,55 +1,95 @@
...
+      this._notifySubscribersUp()
+    } else {
+      this._notifySubscribersDown()
     }
+    this._lastIsRunning = isRunning
+  }
+
+  _notifySubscribersStatus (isRunning: boolean) {
+    this._subscribersStatus.forEach((callback: StatusCallback) => {
+      try { callback(isRunning) } catch (e) {}
```
donce: We shouldn't ignore errors of listeners - we should `logger.info` and `bugReporter.captureInfoException` them.
Same applies for `_notifySubscribersUp` and `_notifySubscribersDown`

src/app/mysterion.js
Waldz: Fixed

src/app/mysterion.js
Waldz: Fixed

src/app/mysterion.js
Waldz: Very exact feedback, fixed

src/app/mysterion.js
Waldz: I need callback, because my intention is to build self explainable code, which makes clear difference between:
- what is being loaded parallely
- what is being loader after process ready

src/libraries/mysterium-client/monitoring.js
Waldz: Not my scope. Fixing in #184

src/libraries/mysterium-client/monitoring.js
Waldz: Fixing in #184

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -18,7 +18,6 @@
 import {Tail} from 'tail'
 import path from 'path'
 import logLevels from '../log-levels'
-import logger from '../../../app/logger'
```
ignasbernotas: why was this removed?

src/renderer/App.vue
```diff
@@ -92,32 +92,26 @@
...
+      this.rendererCommunication.onMysteriumClientDown(() => {
+        this.$store.dispatch('setClientRunningState', false)
 
-        this.$store.dispatch(type.OVERLAY_ERROR, {
-          message: 'mysterium_client is down',
-          hint: 'Please give it a moment to boot. If this message persists try restarting the app or please contact support'
-        })
-        this.$store.dispatch('setClientRunningState', clientRunningState)
+        // do nothing while on terms page
+        if (this.$route.name !== 'terms') {
```
ignasbernotas: I would prefer:
`if (this.$route.name == 'terms') { return; }`

src/libraries/mysterium-client/launch-daemon/process.js
Waldz: Because it's `mysterium-client` library and library should not depend on Mysterion application

src/renderer/App.vue
Waldz: Too small improvement to worry about. Such conditional behaviour should be dropped at all

src/app/mysterion.js
```diff
@@ -232,65 +253,58 @@ class Mysterion {
     } catch (e) {
       const error = new Error(translations.termsAcceptError)
       error.original = e
-      logger.error(error)
       throw error
     }
-
-    this.window.resize(this.config.windows.app)
```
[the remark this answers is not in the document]
Waldz: Moved to upper layer

src/app/mysterion.js
```diff
@@ -189,27 +212,25 @@ class Mysterion {
...
   // declining terms will quit the app
   async _acceptTermsOrQuit () {
+    logInfo('Accepting terms')
     try {
       const accepted = await this._acceptTerms()
       if (!accepted) {
-        logger.info('Terms were refused. Quitting.')
+        logInfo('Terms were refused. Quitting.')
         app.quit()
-        return false
```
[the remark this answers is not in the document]
Waldz: Restored previous code

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -46,7 +49,7 @@ class Process {
 
   async stop () {
     await this.tequilapi.stop()
-    logger.info('Client Quit was successful')
+    console.log('Client Quit was successful')
```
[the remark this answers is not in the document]
Waldz: Refer to https://github.com/MysteriumNetwork/mysterion/pull/161#discussion_r192098875

src/app/mysterion.js
```diff
@@ -232,65 +259,58 @@ class Mysterion {
...
-      try {
-        this.communication.sendHealthCheck({ isRunning: this.monitoring.isRunning() })
-      } catch (e) {
-        this.bugReporter.captureException(e)
-        return
-      }
-
-      setTimeout(() => updateRendererWithHealth(), 1500)
-    }
+  async _startProcess () {
```
donce: `async` is not needed anymore.

Waldz: Removed

PR #163, opened 2018-05-16
Refactor Mysterion bootstrap method

donce:
This bootstrap method is very important, because application won't start if it works incorrectly, but it's very hard to understand what it does - I'm trying to improving readability of it.

### DIFF
--- a/src/app/mysterion.js
+++ b/src/app/mysterion.js
@@ -45,35 +45,74 @@ class Mysterion {
   }
 
   async bootstrap () {
-    let termsAccepted
-    try {
-      this.terms.load()
-      termsAccepted = this.terms.isAccepted()
-    } catch (e) {
-      termsAccepted = false
-      this.bugReporter.captureException(e)
+    const showTerms = !this._AreTermsAccepted()
+    const browserWindow = this._tryCreatingBrowserWindow()
+    const windowSize = this._getWindowSize(showTerms)
+    this.window = this._tryCreatingWindow(windowSize)
+
+    const send = this._getSendFunction(browserWindow)
+    this.messageBus = new MainMessageBus(send, this.bugReporter.captureException)
+    this.communication = new MainCommunication(this.messageBus)
+
+    await this._onRendererLoaded()
+
+    if (showTerms) {
+      const accepted = await this._acceptTermsOrQuit()
+      if (!accepted) {
+        return
+      }
     }
 
-    let browserWindow
+    await this._ensureDaemonInstallation()
+    await this._startProcessAndMonitoring()
+  }
+
+  _getWindowSize (showTerms) {
+    if (showTerms) {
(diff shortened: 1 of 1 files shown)


src/app/mysterion.js
```diff
@@ -45,35 +45,74 @@ class Mysterion {
   }
 
   async bootstrap () {
```
ignasbernotas: I know it's a lot of work, but maybe we can combine forces and refactor it to use flow.

src/app/mysterion.js
```diff
@@ -45,35 +45,74 @@ class Mysterion {
...
+
+  _getWindowSize (showTerms) {
+    if (showTerms) {
+      return this.config.windows.terms
+    } else {
+      return this.config.windows.app
+    }
+  }
+
+  _AreTermsAccepted () {
```
ignasbernotas: "Are" -> "are"

src/app/mysterion.js
```diff
@@ -45,35 +45,74 @@ class Mysterion {
...
+      return false
     }
+  }
+
+  _getSendFunction (browserWindow) {
+    return browserWindow.webContents.send.bind(browserWindow.webContents)
+  }
 
-    let browserWindow
+  _tryCreatingBrowserWindow () {
```
ignasbernotas: I don't think this method needs to be prefixed with try if it still throws an exception on failure. I think createBrowserWindow is sufficient.

src/app/mysterion.js
```diff
@@ -45,35 +45,74 @@ class Mysterion {
...
       console.error(e)
       this.bugReporter.captureException(e)
       throw new Error('Failed to open window.')
     }
+  }
 
-    const send = browserWindow.webContents.send.bind(browserWindow.webContents)
-    this.messageBus = new MainMessageBus(send, this.bugReporter.captureException)
-    this.communication = new MainCommunication(this.messageBus)
+  _tryCreatingWindow (size) {
```
ignasbernotas: creatingWindow and creatingBrowserWindow naming is confusing, these could be merged into one?

src/app/mysterion.js
donce: Totally agree, but let's not increase scope of this PR, it's not healthy. After this is in `master`, we can work on that :)

src/app/mysterion.js
donce: Done

src/app/mysterion.js
donce: Done

src/app/mysterion.js
donce: Renamed to `createWindow` and `createBrowserWindow`. It is confusing, but at least it is consistent - i.e. we have `browserWindowFactory, windowFactory` properties in `Mysterion` class.

PR #164, opened 2018-05-16
MYST-508 Monitor system log file for 'mysterium client' errors

Waldz:


src/app/communication/utils.js
```diff
@@ -1,18 +1,42 @@
...
+ * Subscribes for specific event and resolves when first event is received.
+ *
+ * @param subscriber - function to subscribe for specific event
+ * @param timeout - timeout in miliseccons
+ *
+ * @returns {Promise<any>}
+ */
+function onFirstEventOrTimeout (subscriber: SubscriberCallback, timeout: number): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const timer = setTimeout(
```
shroomist: most common convention that I've seen is the following:
```
const timer = setTimeout(()=> {
    reject(new Error('Promise timed out after ' + timeout + ' ms')
}, timeout)

Comment (Waldz) [reply src/app/communication/utils.js]: If You talk about code alignment, same rules apply as for all code:
1. short statements:
```
 callFunction(() => doNothing(), 100, true, config)
```

1. long statements:
```
 callFunction(
  () => {
    // do
  },
  100,
  true,
  createNewConfig()
)
```

File: src/app/communication/utils.js [hunk 2]
```diff
@@ -1,18 +1,42 @@
 // @flow
 type Callback = (data: any) => void
+type SubscriberCallback = (Callback) => void
```
donce: `SubscriberCallback` is confusing - it sounds like a `callback`, which is a function you're supposed to call after you're done. But it's not a callback, it's a subscriber.

Simply `Subscriber` would cause less confusion.

src/app/communication/utils.js
```diff
@@ -1,18 +1,42 @@
...
+ *
+ * @param subscriber - function to subscribe for specific event
+ * @param timeout - timeout in miliseccons
+ *
+ * @returns {Promise<any>}
+ */
+function onFirstEventOrTimeout (subscriber: SubscriberCallback, timeout: number): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const timer = setTimeout(
+      () => reject(new Error('Promise timed out after ' + timeout + ' ms')),
```
donce: Please use string interpolation:
`Promise timed out after ${timeout} ms`

src/app/mysterion.js
```diff
@@ -211,6 +216,18 @@ class Mysterion {
     this.monitoring.start()
   }
 
+  async waitProcess (callback) {
+    // Block rendering until process is launched
+    try {
+      await onFirstEventOrTimeout(this.monitoring.subscribeUp.bind(this.monitoring), MYSTERIUM_CLIENT_STARTUP_THRESHOLD)
+    } catch (e) {
+      this.communication.sendErrorToRenderer(translations.processStartError)
+      throw new Error("Failed to start 'mysterium_client' process. " + e)
```
donce: By re-throwing error, we are loosing stacktrace here.

src/app/mysterion.js
```diff
@@ -211,6 +216,18 @@ class Mysterion {
     this.monitoring.start()
   }
 
+  async waitProcess (callback) {
```
donce: async with callback? This is confusing.. Why do we need this?

src/libraries/mysterium-client/launch-daemon/config.js
```diff
@@ -0,0 +1,5 @@
+import path from 'path'
+
+export const InverseDomainPackageName = 'network.mysterium.mysteriumclient'
```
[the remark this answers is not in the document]
donce: Or maybe `INVERSE_DOMAIN_PACKAGE_NAME`, since it's a constant?

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -27,40 +30,42 @@ class Process {
...
   onLog (level, cb) {
-    tailFile(this._getFileForLevel(level), cb)
+    if (level === logLevels.LOG) {
+      tailFile(path.join(this.logDirectory, 'stdout.log'), cb)
+      return
+    }
+
+    if (level === logLevels.ERROR) {
+      tailFile(path.join(this.logDirectory, 'stderr.log'), cb)
+      tailFile(SYSTEM_LOG, filterLine(InverseDomainPackageName, cb))
```
donce: What happens if we fail to read system log file? I believe it's very likely, we shouldn't make many expectations about os environment we're running.

src/app/communication/utils.js
Waldz: Renamed

src/app/communication/utils.js
Waldz: Fixed

src/app/mysterion.js
Waldz: I dont need stack-trace from library, I am printing more exact reason

src/libraries/mysterium-client/launch-daemon/config.js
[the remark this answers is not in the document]
Waldz: Fixed. Using uppercase for constants

src/libraries/mysterium-client/launch-daemon/process.js
Waldz: Exactly, this is why this class name is `launch-daemon/process.js`. 
Dont worry, it's loaded only in Darwin

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -27,40 +30,42 @@ class Process {
...
-  try {
-    const logTail = new Tail(filePath)
-    logTail.on('line', cb)
-    logTail.on('error', cb)
-  } catch (e) {
-    console.error('log file watching failed. file probably doesn\'t exist: ' + filePath)
+  const logTail = new Tail(filePath)
+  logTail.on('line', cb)
+  logTail.on('error', () => {
+    console.error(`log file watching failed. file probably doesn't exist: ` + filePath)
```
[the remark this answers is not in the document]
Waldz: Fixed

src/app/mysterion.js
```diff
@@ -246,10 +247,13 @@ class Mysterion {
   /**
    * Blocking function until process is launched
    */
-  async _onProcessReady (callback) {
-    await onFirstEvent(this.monitoring.subscribeUp.bind(this.monitoring))
-
-    callback()
+  _onProcessReady (callback) {
+    onFirstEventOrTimeout(this.monitoring.subscribeUp.bind(this.monitoring), MYSTERIUM_CLIENT_STARTUP_THRESHOLD)
+      .then(() => callback())
```
shroomist: can simplify this to be `.then(callback)`

PR #165, opened 2018-05-16
HOTFIX: Make window rendering flow clearer

Waldz:


src/app/mysterion.js
```diff
@@ -204,7 +204,8 @@ class Mysterion {
       this.communication.sendProposals(await this.proposalFetcher.fetch())
     })
 
-    this.messageBus.send(busMessages.APP_START)
+    console.log('Notify window to finish rendering')
+    this.communication.sendRendererLoadContinue()
```
shroomist: I'd like this event to be more specific 'sendMysteriumClientIsReady'

src/app/communication/renderer-communication.js
```diff
@@ -22,6 +22,14 @@ class RendererCommunication {
 
   // TODO: remaining other messages
 
+  sendRendererLoadStarted (): void {
```
ignasbernotas: It's still not clear what it's for. Need to figure out a better name for it. Something like sendRenderedBootedMessage is more implicit.

src/app/communication/renderer-communication.js
```diff
@@ -22,6 +22,14 @@ class RendererCommunication {
...
+    return this._send(messages.RENDERER_LOAD_STARTED)
+  }
+
+  onRendererLoadContinue (callback: () => void) {
+    this._on(messages.RENDERER_LOAD_CONTINUE, callback)
```
ignasbernotas: Something like onRendererReady is more clearer.

src/app/communication/main-communication.js
```diff
@@ -20,8 +20,19 @@ class MainCommunication {
     this._messageBus = messageBus
   }
 
+  onRendererLoadStarted (callback: () => void) {
```
ignasbernotas: onRendererBooted?

src/app/mysterion.js
Waldz: Renamed

src/app/communication/renderer-communication.js
Waldz: `sendMysteriumClientIsReady` makes more sence about event. Renamed

src/app/communication/main-communication.js
Waldz: Renamed

src/app/communication/renderer-communication.js
ignasbernotas: But it's not the mysterium_client that's ready, it's the whole backend logic that's ready. Once it is ready we need to notify the renderer that it can start it's business logic flow, therefore `sendMysteriumClientIsReady` doesn't explain that.

I prefer:
renderer booted | renderer initializing | renderer started
and
renderer ready
events.

src/app/communication/messages.js
```diff
@@ -4,16 +4,17 @@ export default {
   CONNECTION_REQUEST: 'connection.request',
   CONNECTION_CANCEL: 'connection.cancel',
 
+  MYSTERIUM_CLIENT_READY: 'renderer.load.continue',
```
shroomist: can fix the value of const `renderer.client.ready`

src/app/communication/renderer-communication.js
Waldz: Yes, currently we're unlocking identity in this event. I would like to have smaller event.

Waldz: One single events leads to loading, while some things can be parrelalized

src/app/communication/messages.js
Waldz: fixed

PR #170, opened 2018-05-21
Fix some of Flow warnings

donce:


src/libraries/mysterium-tequilapi/dto/connection-status-enum.js
```diff
@@ -7,7 +7,7 @@ const ConnectionStatusEnum = {
   CONNECTING: 'Connecting'
 }
 
-type ConnectionStatus = 'Connected' | 'Connecting' | 'NotConnected' | 'Connecting'
+type ConnectionStatus = 'Connected' | 'NotConnected' | 'Disconnecting' | 'Connecting'
```
ignasbernotas: oops, good catch

test/unit/renderer/store/modules/tequilapi-manipulator.js
```diff
@@ -69,29 +69,26 @@ function factoryTequilapiManipulator () {
       this.setConnectFail(false)
       this.setConnectFailClosedRequest(false)
     },
-    setStatusFail: function (error: Error) {
-      statusFail = error
+    setStatusFail: function (value: boolean) {
```
ignasbernotas: Does it not expect an error value?

src/libraries/mysterium-tequilapi/dto/connection-status-enum.js
donce: Flow caught it :)
Also, it caught a real bug, not just interface unalligments :)

PR #171, opened 2018-05-21
Feature/mys 2 disconnect notifications

shroomist:
push notifications for when disconnected


src/app/mysterion.js
```diff
@@ -231,6 +242,12 @@ class Mysterion {
     this.communication.onCurrentIdentityChange((identity) => {
       this.bugReporter.setUser(identity)
     })
+
+    this.communication.onConnectionStatusChange((status) => {
+      if (this.disconnectNotification.isEnabled && status.newStatus === ConnectionStatusEnum.NOT_CONNECTED) {
```
ignasbernotas: `this.disconnectNotification.isEnabled` could be isolated in the `show` method, don't see why we need it outside of the class scope:)

src/app/notification/disconnect.js
```diff
@@ -0,0 +1,29 @@
+// @flow
```
ignasbernotas: The file name is `disconnect.js`, class is `MystNoticiation`, imported as `DisconnectNotification`. I would make this a generic notification class (`notification.js` since it already is, this code has nothing to do with disconnect).

src/app/userSettings.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+import {readFileSync, writeFileSync} from 'fs'
+
+export default class UserSettings {
```
ignasbernotas: User settings exposes as a single mutable property. I would change this to something along the lines:
```js
class UserSettings {
  _settingsPath: string
  _showDisconnectNotifications: boolean = true

  constructor (settingsPath: string) {
    this._settingsPath = settingsPath
  }

  load () {
    const settings = loadSettings(this._settingsPath)

    this._showDisconnectNotifications = settings.showDisconnectNotifications
    // same for others
  }

  enableDisconnectNotifications () {
    this._showDisconnectNotifications = true
  }

  disableDisconnectNotifications () {
    this._showDisconnectNotifications = false
  }

  disconnectNotificationsEnabled () {
    return this._showDisconnectNotifications
  }

  save () {
    saveSettings(this._settingsPath, {
      showDisconnectNotifications: this._showDisconnectNotifications
    })
  }
}
```

Clearer API and easier to build (into DI containter) and use:)

src/app/notification/disconnect.js
```diff
@@ -0,0 +1,29 @@
...
+
+  enable () {
+    this.isEnabled = true
+  }
+
+  disable () {
+    this.isEnabled = false
+  }
+
+  show () {
```
ignasbernotas: Don't see why we need public properties, if(!this.isEnabled) return;

src/app/notification/disconnect.js
```diff
@@ -0,0 +1,29 @@
+// @flow
+import {Notification} from 'electron'
+
+export default class MystNotification {
+  title: string
```
ignasbernotas: these should be prefixed with _

src/app/userSettings.js
shroomist: we've discussed this with @donce, his suggestion was to keep single responsibility class.

src/app/notification/disconnect.js
shroomist: I'd have to rename this to `showIfEnabled` or something in that case..

ignasbernotas: I think the name is fine either way, but just don't see why we need to expose these since it's only used for that single show method. @donce what do you think?

src/app/mysterion.js
donce: I also thought about that, but it would mean that `disconnectNotification.show()` would sometimes show notification, sometimes not, depending on internal mutable state - which isn't very readable.

`isEnabled` property doesn't seem to belong to `MystNotification` class - if we extract it out, `MystNotification` would be left with one responsibility - to show notification.

src/app/notification/disconnect.js
donce: I think you both have good points `isEnabled` seems strange in this class, but If we make this method show only when it is enabled, than `show()` name wouldn't be appropriate - it wouldn't show in all cases, so method name would lie.

Previous comment applies here as well:
>  `isEnabled` property doesn't seem to belong to `MystNotification` class - if we extract it out, `MystNotification` would be left with one responsibility - to show notification.

donce: Or we could make it `showIfEnabled`, which would mean 2 responsibilities.

src/app/userSettings.js
donce: @ignasbernotas the problem is that this wouldn't be `UserSettings` class anymore - this would be more like `UserSettingsManager`, because it would have many responsibilities:
- store settings
- save settings
- load settings
- update settings

The biggest problem would be that all users of this class should be aware that any of settings can change at any time, which is very easy to forget.

donce: If you want to have a class for all these responsibilities in one place, maybe it's worth having 2 classes - `UserSettings` for settings values (DTO) and `UserSettingsManager` (any better name?) for saving and loading? In that case, `UserSettingsManager` would have single mutable property (dto) instead of all settings.

src/app/userSettings.js
```diff
@@ -0,0 +1,25 @@
...
+
+function saveSettings (path: string, settings: UserSettings) {
+  const settingsString = JSON.stringify(settings)
+  writeFileSync(path, settingsString)
+}
+
+function loadSettings (path: string): UserSettings {
+  const data = readFileSync(path, {encoding: 'utf8'})
+  if (data !== 'undefined') {
+    return new UserSettings(JSON.parse(data))
```
donce: constructor of `UserSettings` requires boolean field `showDisconnectNotifications` - we're not sure that this field is present in this line, we should check that.

src/app/userSettings.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+import {readFileSync, writeFileSync} from 'fs'
+
+export default class UserSettings {
+  showDisconnectNotifications: boolean
+
+  constructor (obj: { showDisconnectNotifications: boolean }) {
+    this.showDisconnectNotifications = obj.showDisconnectNotifications
+  }
+}
```
donce: Simple `type` instead of class should serve the same purpose:
```js
type UserSettings = {
  showDisconnectNotifications: boolean
}
```

src/app/userSettings.js
```diff
@@ -0,0 +1,25 @@
...
+
+function saveSettings (path: string, settings: UserSettings) {
```
donce: We can specify that this returns `void`.

src/main/dependencies/modules/disconnect-notification.js
```diff
@@ -0,0 +1,20 @@
...
+
+import type {Container} from '../../../app/di'
+import DisconnectNotification from '../../../app/notification/disconnect'
+
+function bootstrap (container: Container) {
+  container.factory(
+    'disconnectNotification',
+    ['userSettings'],
+    (userSettings) => {
+      const disconnectNotification = new DisconnectNotification('Disconnected', 'You have been disconnected from VPN server')
```
donce: `DisconnectNotification` looks like a new class, but in fact it is `MystNotification` class.
We shouldn't rename function here, variable name represents it's purpose well enough:
```js
const disconnectNotification = new MystNotification('Disconnected', 'You have been disconnected from VPN server')
```

src/main/dependencies/modules/user-settings.js
```diff
@@ -0,0 +1,33 @@
...
+    ['userSettingsPath'],
+    (userSettingsPath) => {
+      let userSettings
+      try {
+        userSettings = loadSettings(userSettingsPath)
+      } catch (e) {
+        console.log(`Failing to load ${userSettingsPath}`)
+        console.log(e)
+      }
+      return userSettings
```
donce: No need to have `userSettings` variable:
```js
try {
  return loadSettings(userSettingsPath)
} catch (e) {
  console.log(`Failing to load ${userSettingsPath}`)
  console.log(e)
  return null
}
```

test/helpers/utils.js
```diff
@@ -19,6 +19,14 @@ async function capturePromiseError (promise: Promise<any>): Promise<?Error> {
   return null
 }
 
+function captureError (fn: Function): ?Error {
```
donce: We should specify that this function should not expect any params:
```js
function captureError (fn: () => any): ?Error
```

test/unit/app/userSettings.spec.js
```diff
@@ -0,0 +1,58 @@
...
+import {join} from 'path'
+import {readFileSync, writeFileSync, unlinkSync} from 'fs'
+import {captureError} from '../../helpers/utils'
+
+describe('UserSettings', () => {
+  const settingsPath = join(tmpdir(), 'settings.test.saving.json')
+  const loadSettingsPath = join(tmpdir(), 'settings.test.loading.json')
+  const invalidPath = join(tmpdir(), 'some')
+
+  describe('saves settings', () => {
```
donce: Structure of specs are different from actual code structure - i.e. `saves settings` is behaviour of `saveSettings` function, not of `UserSettings` class.

donce: `describe` should indicate object you're describing here, not a behaviour.

test/unit/app/userSettings.spec.js
```diff
@@ -0,0 +1,58 @@
...
+    })
+
+    it('exports a valid json on save()', () => {
+      saveSettings(settingsPath, userSettings)
+      const data = readFileSync(settingsPath, {encoding: 'utf8'})
+      expect(data.toString()).to.eql('{"showDisconnectNotifications":false}')
+    })
+  })
+
+  describe('loads settings from file', () => {
```
donce: `describe` should indicate object you're describing here, not a behaviour.

test/unit/app/userSettings.spec.js
```diff
@@ -0,0 +1,58 @@
...
+      writeFileSync(
+        loadSettingsPath,
+        JSON.stringify({showDisconnectNotifications: false})
+      )
+    })
+    after(() => {
+      unlinkSync(loadSettingsPath)
+    })
+
+    it('reads showDiscinnectNotifications correctly from file', () => {
```
donce: mistype in `showDiscinnectNotifications`

test/unit/app/userSettings.spec.js
```diff
@@ -0,0 +1,58 @@
...
+      const loadedObject = loadSettings(loadSettingsPath)
+      const loadedUserSettings = new UserSettings(loadedObject)
```
donce: `loadSettings` is returning `UserSettings` itself, no need to create a second instance of `UserSettings`.

test/unit/app/userSettings.spec.js
```diff
@@ -0,0 +1,58 @@
...
+      expect(loadedUserSettings.showDisconnectNotifications).to.be.false
+    })
+  })
+
+  describe('invalid path', () => {
+    it('throws error on save()', () => {
```
donce: This is a behaviour of `saveSettings` - it should be inside there, i.e.
```
describe('saveSettings', () => {
  it('reads showDiscinnectNotifications correctly from file', () => {
    // ...
  })

  it('throws error with invalid path`, () => {
    // ...
  })
})
```

src/app/mysterion.js
```diff
@@ -40,6 +44,13 @@ class Mysterion {
       this.window.show()
     })
     app.on('before-quit', () => {
+      try {
+        saveSettings(this.userSettingsPath)
```
donce: Settings should be saved after they are changed - application could be terminated, so `before-quit` could be skipped. In that case, settings would not be saved.

donce: This is missing second param - you have to specify settings you're saving.

src/app/userSettings.js
```diff
@@ -0,0 +1,25 @@
...
+  const settingsString = JSON.stringify(settings)
+  writeFileSync(path, settingsString)
+}
+
+function loadSettings (path: string): UserSettings {
+  const data = readFileSync(path, {encoding: 'utf8'})
+  if (data !== 'undefined') {
```
donce: I believe it's safe to compare it to `null`:
```js
if (data != null) {
  ...
}
```
Ref: https://flow.org/en/docs/types/maybe/#toc-refining-maybe-types

test/unit/app/userSettings.spec.js
```diff
@@ -0,0 +1,58 @@
...
+import {readFileSync, writeFileSync, unlinkSync} from 'fs'
+import {captureError} from '../../helpers/utils'
+
+describe('UserSettings', () => {
+  const settingsPath = join(tmpdir(), 'settings.test.saving.json')
+  const loadSettingsPath = join(tmpdir(), 'settings.test.loading.json')
+  const invalidPath = join(tmpdir(), 'some')
+
+  describe('saves settings', () => {
+    const userSettings = new UserSettings(settingsPath)
```
donce: This is outdated.

src/app/notification/disconnect.js
```diff
@@ -0,0 +1,29 @@
+// @flow
+import {Notification} from 'electron'
+
+export default class MystNotification {
```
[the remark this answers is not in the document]
shroomist: fixed

src/app/mysterion.js
shroomist: fixed

src/app/notification/disconnect.js
shroomist: fixed

shroomist:
@donce @ignasbernotas comments fixed, please review

test/unit/app/userSettings.spec.js
```diff
@@ -0,0 +1,94 @@
...
+      await userSettingsStore.load()
+
+      expect(userSettingsStore.settings).to.be.eql({showDisconnectNotifications: false})
+    })
+
+    it('throws error if load() fails on invalid path to file', async () => {
+      const userSettingsStore = new UserSettingsStore(invalidPath)
+      const error = await capturePromiseError(userSettingsStore.load())
+
+      expect(error instanceof Error).to.be.true
```
donce:
```js
expect(error).to.be.an.instanceof(Error)
```

test/unit/app/userSettings.spec.js
```diff
@@ -0,0 +1,94 @@
...
+describe('UserSettingsStore', () => {
+  describe('save()', () => {
+    const saveSettingsPath = join(tmpdir(), 'settings.test.saving.json')
+    const invalidPath = join(tmpdir(), 'some', 'dir')
+
+    after(() => {
+      unlinkSync(saveSettingsPath)
+    })
+
+    it('exports a valid json on save()', async () => {
```
donce: `'on save()'` at the end of description is redundant - since you're in `describe('save()'`, you're already describing behaviour of `save()`

src/app/mysterion.js
```diff
@@ -80,6 +83,11 @@ class Mysterion {
     this.communication = new MainCommunication(this.messageBus)
 
     await this._onRendererLoaded()
+    try {
```
ignasbernotas: I would move this into a separate method like `loadUserSettings`

src/app/mysterion.js
```diff
@@ -80,6 +83,11 @@ class Mysterion {
...
+      await this.userSettingsStore.load()
+    } catch (e) {
+      this.userSettingsStore.set({showDisconnectNotifications: true})
```
ignasbernotas: I don't like that mysterion class has to know about showDisconnectNotifications property, I'd make a setDefaults() method on the store class. We need to try to move away from externally mutable properties.

src/app/mysterion.js
```diff
@@ -283,4 +293,22 @@ class Mysterion {
   }
 }
 
+function registerUserSettingsCommunication (userSettingsStore, communication, disconnectNotification) {
```
ignasbernotas: Function name is false, you're doing more than just registering communication. I would split this function into two:
`showNotificationOnDisconnect` and `synchronizeUserSettings`

src/app/communication/main-communication.js
```diff
@@ -108,6 +113,14 @@ class MainCommunication {
     this._on(messages.TERMS_ANSWERED, callback)
   }
 
+  onUserSettingsUpdate (callback: (UserSettings) => void) {
+    this._on(messages.USER_SETTINGS_SET, callback)
+  }
+
+  onUserSettingsRequest (callback: () => void) {
+    this._on(messages.USER_SETTINGS_GET, callback)
```
donce: Why do we have 2 different names for the same thing:
- "User settings request" VS "user settings get"
- "User settings update" VS "user settings set"

Wouldn't it be easier to name them the same, i.e. `onUserSettingsRequest` method and constant?

src/app/mysterion.js
```diff
@@ -163,6 +168,14 @@ class Mysterion {
     }
   }
 
+  async _loadUserSettings () {
+    try {
+      await this.userSettingsStore.load()
+    } catch (e) {
+      this.userSettingsStore.setDefault()
```
donce: This error might not be what we are expecting - we should send it to Sentry as INFO exception.

src/app/mysterion.js
```diff
@@ -284,4 +300,24 @@ class Mysterion {
...
+function showNotificationOnDisconnect (userSettingsStore, communication, disconnectNotification) {
+  communication.onConnectionStatusChange(async (status) => {
+    if (userSettingsStore.get().showDisconnectNotifications &&
+      status.newStatus === ConnectionStatusEnum.NOT_CONNECTED) {
+      disconnectNotification.show()
+    }
+  })
+}
+
+function synchronizeUserSettings (userSettingsStore, communication) {
```
donce: Nice abstraction 👍

src/app/mysterion.js
```diff
@@ -284,4 +300,24 @@ class Mysterion {
   }
 }
 
+function showNotificationOnDisconnect (userSettingsStore, communication, disconnectNotification) {
+  communication.onConnectionStatusChange(async (status) => {
+    if (userSettingsStore.get().showDisconnectNotifications &&
+      status.newStatus === ConnectionStatusEnum.NOT_CONNECTED) {
```
donce: How are we sure that we won't trigger notification when we launch the app, when we are in `NOT_CONNECTED` state?

Also, should we show notification in such case:
- User opens Mysterion - status `NOT_CONNECTED`
- User connects to node - status `CONNECTING`
- Connection fails - status `NOT_CONNECTED`. This will trigger "disconnect" notification, even though nothing was "disconnected".

src/app/user-settings/user-settings-store.js
```diff
@@ -0,0 +1,74 @@
...
+  const data = await readFileAsync(path, {encoding: 'utf8'})
+  const parsedSettings = JSON.parse(data)
+
+  if (typeof parsedSettings.showDisconnectNotifications !== 'boolean') {
+    return null
+  }
+  return parsedSettings
+}
+
+class UserSettingsStore {
```
donce: This class should be before `saveSettings` and `loadSettings`, because this class is top-level object - other methods are just "private" functions.

src/app/user-settings/user-settings-store.js
```diff
@@ -0,0 +1,74 @@
...
+  const settingsString = JSON.stringify(settings)
+  await writeFileAsync(path, settingsString)
+}
+
+async function loadSettings (path: string): Promise<?UserSettings> {
+  const data = await readFileAsync(path, {encoding: 'utf8'})
+  const parsedSettings = JSON.parse(data)
+
+  if (typeof parsedSettings.showDisconnectNotifications !== 'boolean') {
+    return null
```
donce: We can just throw an error in this case - i.e. if JSON parsing fails in this method, we throw error, so this behaviour should be consistent.

src/app/user-settings/user-settings-store.js
```diff
@@ -0,0 +1,74 @@
...
+import type {UserSettings} from './user-settings'
+
+const readFileAsync = promisify(readFile)
+const writeFileAsync = promisify(writeFile)
+
+const defaultSettings: UserSettings = {
+  showDisconnectNotifications: true
+}
+
+async function saveSettings (path: string, settings: UserSettings): Promise<?Error> {
```
donce: It doesn't seem to return anything, why is response type `?Error`?

src/app/user-settings/user-settings-store.js
```diff
@@ -0,0 +1,74 @@
...
+  const parsedSettings = JSON.parse(data)
+
+  if (typeof parsedSettings.showDisconnectNotifications !== 'boolean') {
+    return null
+  }
+  return parsedSettings
+}
+
+class UserSettingsStore {
+  _settings: ?UserSettings
```
donce: We shouldn't make this param `maybe` type - it should always have some value.
I.e. if loading of settings fails, we should keep the defaults.
This will make usage of settings much easier.

src/app/user-settings/user-settings-store.js
```diff
@@ -0,0 +1,74 @@
...
+  constructor (path: string) {
+    this._path = path
+  }
+
+  setDefault () {
+    this.settings = defaultSettings
+  }
+
+  async load (): Promise<?UserSettings> {
+    this._settings = await loadSettings(this._path)
```
donce: You're not returning settings, even though you declare it should return it.

src/app/user-settings/user-settings-store.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  async save (): Promise<?Error> {
+    if (!this._settings) throw new Error('Trying to save UserSettings, but UserSettingsStore.settings is missing')
+    return saveSettings(this._path, this._settings)
```
donce: `saveSettings` doesn't seem to be actually returning anything.

src/app/user-settings/user-settings-store.js
shroomist: fixed

src/app/user-settings/user-settings-store.js
shroomist: done

src/app/user-settings/user-settings-store.js
shroomist: ok

src/app/user-settings/user-settings-store.js
shroomist: good catch

src/app/user-settings/user-settings-store.js
shroomist: right on

src/app/mysterion.js
```diff
@@ -284,4 +300,31 @@ class Mysterion {
   }
 }
 
+function showNotificationOnDisconnect (userSettingsStore, communication, disconnectNotification) {
+  communication.onConnectionStatusChange(async (status) => {
+    const shouldShowNotification =
+      userSettingsStore.get().showDisconnectNotifications &&
+      (status.newStatus === ConnectionStatusEnum.NOT_CONNECTED &&
+        (status.oldStatus === ConnectionStatusEnum.DISCONNECTING ||
```
donce: If we remove this single line, I think we'll achieve what we initially wanted without much effort - we would show notification only when disconnected unexpectedly, without user pressing "disconnect" button.

When user is disconnecting, states go "CONNECTED" -> "DISCONNECTING" -> "NOT CONNECTED". When disconnect happens unexpectedly, we go from "CONNECTED" to "NOT CONNECTED" without "DISCONNECTING" state.

src/app/user-settings/user-settings-store.js
```diff
@@ -0,0 +1,87 @@
...
+  async save (): Promise<void> {
+    if (!this._settings) throw new Error('Trying to save UserSettings, but UserSettingsStore.settings is missing')
+    return saveSettings(this._path, this._settings)
+  }
+
+  set (settings: UserSettings) {
+    this._settings = settings
+  }
+
+  get (): ?UserSettings {
```
donce: return type is `UserSettings`.

src/app/user-settings/user-settings-store.js
```diff
@@ -0,0 +1,87 @@
...
+    } catch (e) {
+      if (isFileNotExistError(e)) {
+        return
+      }
+      throw e
+    }
+  }
+
+  async save (): Promise<void> {
+    if (!this._settings) throw new Error('Trying to save UserSettings, but UserSettingsStore.settings is missing')
```
donce: This conditional is redundant, `this._settings` cannot be null.

shroomist: fixed

src/app/mysterion.js
shroomist: ok

src/app/user-settings/user-settings-store.js
shroomist: yes

PR #172, opened 2018-05-22
Fix Flow warnings about TequilapiClient mocks

donce:
Added `TequilapiClient` interface with `HttpTequilapiClient` real implementation and other mock implementations.

### DIFF
--- a/test/unit/renderer/store/modules/connection.spec.js
+++ b/test/unit/renderer/store/modules/connection.spec.js
@@ -92,10 +92,10 @@ function factoryTequilapiManipulator () {
   }
 
   return {
-    getFakeApi: function () {
+    getFakeApi () {
       return new ConnectionApiManipulator()
     },
-    cleanup: function () {
+    cleanup () {
       this.setStatusFail(false)
       this.setStatisticsFail(false)
       this.setIpFail(false)
@@ -103,25 +103,25 @@ function factoryTequilapiManipulator () {
       this.setConnectFail(false)
       this.setConnectFailClosedRequest(false)
     },
-    setStatusFail: function (value: boolean) {
+    setStatusFail (value: boolean) {
       statusFail = value
     },
-    setStatisticsFail: function (value: boolean) {
+    setStatisticsFail (value: boolean) {
       statisticsFail = value
     },
-    setIpTimeout: function (value: boolean) {
+    setIpTimeout (value: boolean) {
       ipTimeout = value
     },
-    setIpFail: function (value: boolean) {
+    setIpFail (value: boolean) {
       ipFail = value
     },
-    setConnectFail: function (value: boolean) {
+    setConnectFail (value: boolean) {
       connectFail = value
     },
-    setConnectFailClosedRequest: function (value: boolean) {
+    setConnectFailClosedRequest (value: boolean) {
       connectFailClosedRequest = value
     },
-    getFakeError: function (): Error {
+    getFakeError (): Error {
       return errorMock
     }
   }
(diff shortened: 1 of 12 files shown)


test/unit/renderer/store/modules/connection.spec.js
```diff
@@ -32,6 +31,115 @@ import type {EventFactory as StatsEventsFactory} from '../../../../../src/app/st
...
+    async connectionStatistics (): Promise<ConnectionStatisticsDTO> {
+      if (statisticsFail) {
+        throw errorMock
+      }
+      return new ConnectionStatisticsDTO({duration: 1})
+    }
+  }
+
+  return {
+    getFakeApi: function () {
```
shroomist: you can use shorthand syntax
```
getFakeApi () {
...
},
cleanup () {
...
}```

File: test/unit/renderer/store/modules/connection.spec.js [hunk 2]
```diff
@@ -32,6 +31,115 @@ import type {EventFactory as StatsEventsFactory} from '../../../../../src/app/st
...
+  let ipFail = false
+  let ipTimeout = false
+  let connectFail = false
+  let connectFailClosedRequest = false
+
+  const errorMock = new Error('Mock error')
+  const timeoutErrorMock = createMockTimeoutError()
+  const closedRequestErrorMock = createMockRequestClosedError()
+
+  class ConnectionApiManipulator extends TequilapiManipulator {
```
shroomist: looks like you're declaring a class inside a factory function O_O

shroomist: also this class perhaps should be appended with `Mock` as its in tests

ignasbernotas: I assume he's doing it since ConnectionApiManipulator uses local vars, and the outer scope doesn't really need anything else from it

donce: Yes, @ignasbernotas is correct. This is not ideal, but previously the same logic was implemented by constructing object, which is passed in as `TequilapiClient`, so extracting that logic into a class which implements interface looks like a first towards perfection.

test/unit/renderer/store/modules/connection.spec.js
donce: Changed.

test/unit/renderer/store/modules/connection.spec.js
donce: Unified mock classes naming with `Mock` postfix.

donce:
Comments fixed, ready for review :)

test/unit/renderer/store/modules/connection.spec.js
```diff
@@ -32,6 +31,115 @@ import type {EventFactory as StatsEventsFactory} from '../../../../../src/app/st
 import {ActionLooper, ActionLooperConfig} from '../../../../../src/renderer/store/modules/connection'
 import ConnectionStatisticsDTO from '../../../../../src/libraries/mysterium-tequilapi/dto/connection-statistics'
 import type {BugReporter} from '../../../../../src/app/bug-reporting/interface'
+import EmptyTequilapiClientMock from './empty-tequilapi-client-mock'
+import ConnectionStatusDTO from '../../../../../src/libraries/mysterium-tequilapi/dto/connection-status'
+import ConnectionIPDTO from '../../../../../src/libraries/mysterium-tequilapi/dto/connection-ip'
+
+function factoryTequilapiManipulator () {
```
shroomist: I don't like the naming here.
suggestion: `constructTequilapiTestHelper`

PR #173, opened 2018-05-22
feature/MYS-2 use DI for rendererCommunication

shroomist:
rendererCommunication will be used for userSettings interactions, so pulling this out into DI


src/renderer/dependencies/modules/communication.js
```diff
@@ -0,0 +1,33 @@
...
+import type {Container} from '../../../app/di'
+import RendererMessageBus from '../../../app/communication/rendererMessageBus'
+import RendererCommunication from '../../../app/communication/renderer-communication'
+
+export default function bootstrap (container: Container) {
+  container.constant('rendererMessageBus', new RendererMessageBus())
+
+  container.factory(
+    'rendererCommunication',
+    ['rendererMessageBus'],
```
donce: This is already implemented:
https://github.com/MysteriumNetwork/mysterion/blob/master/src/renderer/dependencies/modules/application.js#L34

shroomist: ok, fixed

PR #174, opened 2018-05-23
MYS-38 Add INFO level for known captured exceptions

donce:


src/app/bug-reporting/feedback-form.js
```diff
@@ -26,7 +26,7 @@ class FeedbackForm {
   }
 
   show () {
-    this.raven.captureMessage('User opened issue report form.')
+    this.raven.captureMessage('User opened issue report form.', { level: 'info' })
```
shroomist: are you sure we need {level} here? I think it's default for `captureMessage`.
If it's not, can you make it so?

donce: Yes, we need `{level}` here, because `error` is the default - and in Sentry it's shown as an error.

What do you mean making a default? We are using `Raven` directly here, not through `bugReporter`.

PR #175, opened 2018-05-23
Add Logger class, use it in Mysterion class

donce:
What do you think about such abstraction?
Few of advantages:
- Future possibility to filter outputted logging levels (easier to find important errors)
- Future possibility to add more levels
- Single place for changing logging format, i.e. including timestamp
- Clear, consistent interface - currently we have a mess around logging
- Future possibility to merge `Logger` and `BugReporting` into one - i.e. Logger.error could send error to sentry.

If you approve this approach, I'll convert other logging usages to this as well :)

### DIFF
--- a/src/app/logger.js
+++ b/src/app/logger.js
@@ -19,11 +19,11 @@
 
 class Logger {
   info (...data: Array<any>): void {
-    console.info(data)
+    console.info(...data)
   }
 
   error (...data: Array<any>): void {
-    console.error(data)
+    console.error(...data)
   }
 }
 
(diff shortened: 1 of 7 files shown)


donce:
So we though of 3 possibilities for `Logger` initialisation:
- `export default new Logger()` (as it's currently in PR)
- `export default { error, info }`
- use DI for logger

src/app/logger.js
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// @flow
+
+class Logger {
+  info (...data: Array<any>): void {
+    console.info(data)
```
shroomist: please do `console.info(...data)` to spread the the packed array back to individual parameters

donce: Fixed.

PR #182, opened 2018-05-30
Convert Mysterion to Flow

donce:


src/app/mysterion.js
```diff
@@ -15,35 +15,78 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import {app} from 'electron'
+// @flow
+
+import { app, BrowserWindow } from 'electron'
 import trayFactory from '../main/tray/factory'
-import {logLevels as processLogLevels} from '../libraries/mysterium-client'
+import { Installer, logLevels as processLogLevels } from '../libraries/mysterium-client'
```
shroomist: spaces in curlies - I find that  @ignasbernotas has different config for this.
I like it this way better, it also makes navigation a bit easier.

src/app/mysterion.js
```diff
@@ -146,7 +190,7 @@ class Mysterion {
     }
   }
 
-  _getSendFunction (browserWindow) {
+  _getSendFunction (browserWindow: Object) {
```
shroomist: (browserWindow: BrowserWindow)

src/app/mysterion.js
```diff
@@ -291,7 +336,7 @@ class Mysterion {
     this.monitoring.start()
   }
 
-  _onProcessReady (callback) {
+  _onProcessReady (callback: Function) {
```
shroomist: let's consolidate this to (callback: () => void)

src/app/mysterion.js
ignasbernotas: 👍 fixed it on my end, thanks for noticing:)

src/app/mysterion.js
donce: Fixed.

src/app/mysterion.js
donce: Fixed.

PR #186, opened 2018-06-05
Improved tray proposal sync

ignasbernotas:
Tray proposal sync is now more user friendly and responsive. If there are no proposals - a "No countries available" menu item is shown. Also updates tray items not only periodically but also when tray closes.


src/app/communication/main-communication.js
```diff
@@ -60,7 +60,7 @@ interface MainCommunication {
 
   onCurrentIdentityChange (callback: (CurrentIdentityChangeDTO) => void): void,
 
-  onProposalUpdateRequest (callback: () => void): void,
+  onProposalUpdateRequest (callback: () => void | Promise<void>): void,
```
ignasbernotas: Had to change this to allow async callbacks

test/helpers/fakeMessageBus.js
```diff
@@ -18,7 +18,7 @@
 // @flow
 import type {MessageBus} from '../../src/app/communication/messageBus'
 
-type Callback = (data?: mixed) => void
+type Callback = (data?: mixed) => void | Promise<void>
```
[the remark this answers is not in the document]
ignasbernotas: I think both are valid, mine is just a bit shorter (tested both and they seemed to work from IDE perspective)

src/main/tray/factory.js
```diff
@@ -49,6 +49,9 @@ const trayFactory = (
   tray.build()
 
   communication.onConnectionStatusChange((change: ConnectionStatusChangeDTO) => tray.setStatus(change.newStatus))
+  communication.onProposalUpdateRequest(async () => {
+    tray.setProposals(await proposalFetcher.fetch())
```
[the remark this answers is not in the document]
ignasbernotas: Cool, this will reduce the number of required changes, cheers!

src/app/communication/main-communication.js
ignasbernotas: Change removed.

src/main/tray/factory.js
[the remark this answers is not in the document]
ignasbernotas: Fixed.

test/helpers/fakeMessageBus.js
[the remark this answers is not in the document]
ignasbernotas: Change removed.

src/main/tray/factory.js
```diff
@@ -49,6 +49,9 @@ const trayFactory = (
   tray.build()
 
   communication.onConnectionStatusChange((change: ConnectionStatusChangeDTO) => tray.setStatus(change.newStatus))
+  communication.onProposalUpdateRequest(() => {
+    proposalFetcher.fetch().then((proposals) => tray.setProposals(proposals))
+  })
```
donce: Why do we need fetching proposals from 2 places - here and in `mysterion.js`?
```js
    this.communication.onProposalUpdateRequest(async () => {
      this.communication.sendProposals(await this.proposalFetcher.fetch())
    })
```

src/main/tray/translations.js
```diff
@@ -25,7 +25,8 @@ const translations = {
   statusDisconnected: 'Status: Disconnected',
   statusConnected: 'Status: Connected',
   statusConnecting: 'Status: Connecting',
-  statusDisconnecting: 'Status: Disconnecting'
+  statusDisconnecting: 'Status: Disconnecting',
+  noCountriesAvailable: 'No countries available'
```
donce: We already have this message in `app/messages.js`, maybe it's worth re-using that one?

test/unit/main/tray/tray.spec.js
```diff
@@ -170,7 +173,9 @@ describe('tray', () => {
         tray.build()
         tray.setProposals([])
 
-        expect(calledUpdateProposals).to.equal(false)
+        // if calledSetContextMenu is larger than 1 it means tray._update() was called
```
donce: did you meant *... larger than 0...*?

test/unit/main/tray/tray.spec.js
```diff
@@ -170,7 +173,9 @@ describe('tray', () => {
...
+        expect(calledSetContextMenu).to.equal(1)
+        expect(calledUpdateProposals).to.equal(true)
```
donce: We should also check `it('doesn\'t rerender tray` - this can be done easily by adding mock method `build()` on `menuItemBuilder` and checking that it was not called.

ignasbernotas: `menuItemBuilder.build()` isn't that important to us since it doesn't re-render tray, it just maps internal tray menu objects into electron menu objects and returns them, the real rendering is done in `_electronTray.setContextMenu` which is covered in this test

src/main/tray/factory.js
ignasbernotas: Fixed

src/main/tray/translations.js
ignasbernotas: fixed

src/main/tray/tray.js
```diff
@@ -111,6 +112,7 @@ class Tray {
     this._electronTray.on('click', handler)
     this._electronTray.on('mouse-enter', handler)
     this._electronTray.on('mouse-leave', () => {
+      this._update()
```
shroomist: can we do `this._update()` inside `this._electronTray.on('click', handler)`?
it might slow down responsiveness a little bit, but this way we would always be building submenu according to the latest proposals.

ignasbernotas: No, if we do it there, it will close the opened tray :/

PR #189, opened 2018-06-05
Update loading time

mipo47:
MYS-53


test/unit/renderer/pages/VpnLoader.spec.js
```diff
@@ -81,7 +81,6 @@ describe('VpnLoader', () => {
   async function mountAndPrepareLoadingScreen (tequilapi: TequilapiClient, vpnInitializer: Object) {
     const vm = await mountComponent(tequilapi, vpnInitializer)
     await nextTick() // wait for delay inside loader callback
```
donce: I think this `nextTick()` is not needed anymore as well.

donce:
For branch names, we use pattern: `type/MYS-xx-feature-name`
Type can be `feature`, `hotfix`, `fix`, `refactor`

I.e. for this PR, something like this would work: `feature/MYS-53-remove-loading-sleep`

But it's fine for this PR, just have in mind for next PRs :)

shroomist:
this doesn't seem to be solving MYS-53 story, just removes the timers.
If you're suggesting an alternative solution for the story, please add your arguments in the PR description, why do you do it the other way than described in the ticket.

PR #190, opened 2018-06-05
hotfix. new identity is created each time Mysterion is run

shroomist:
identitiesList response is not an array of identities but
`response = {identities: [IdentityDTO{id: 0xBEEF}]}`

### DIFF
--- a/src/libraries/mysterium-tequilapi/dto/identities-response.js
+++ b/src/libraries/mysterium-tequilapi/dto/identities-response.js
@@ -18,14 +18,14 @@
 // @flow
 import IdentityDTO from './identity'
 
-type ResponseMap = Array<Object>
+type IdentityListResponse = { identities: Array<Object> }
 
 class IdentitiesResponseDTO {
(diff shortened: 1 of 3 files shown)


src/libraries/mysterium-tequilapi/dto/identities-response.js
```diff
@@ -18,14 +18,14 @@
...
+type IdentityListResponse = { identities: Array<Object> }
 
 class IdentitiesResponseDTO {
   identities: Array<IdentityDTO>
 
-  constructor (responseData: ResponseMap) {
-    if (typeof responseData !== 'undefined' && Array.isArray(responseData)) {
-      this.identities = responseData.map((identity) => new IdentityDTO(identity))
+  constructor (responseData: IdentityListResponse) {
+    if (responseData && responseData.identities && Array.isArray(responseData.identities)) {
```
donce: `responseData.identities` check is actually not needed, because it's already checked by `Array.isArray(responseData.identities)` - `null` is not an array as well.

PR #191, opened 2018-06-06
Bugfix/mys 75 log file creation

shroomist:
create empty log files on installation. this fixes app freeze in initialisation step for fresh installs

### DIFF
--- a/src/app/messages.js
+++ b/src/app/messages.js
@@ -22,7 +22,8 @@ export default {
   initializationError: {
     message: 'Failed to load Mysterion'
   },
-  processInstallationError: 'Failed to install mysterium_client daemon. Please restart the app and grant permissions.',
+  processInstallationPermissionsError: 'Failed to install Mysterion. Please restart the app and grant permissions.',
+  processInstallationError: 'Failed to install Mysterion.',
   processStartError: 'Failed to start mysterium_client daemon. Please restart the app and try again.',
   termsAcceptError: 'Failed to make a local copy of terms and conditions. Please restart the app and try again.'
 }
--- a/src/app/mysterion.js
+++ b/src/app/mysterion.js
@@ -234,8 +234,12 @@ class Mysterion {
       try {
         await this.installer.install()
       } catch (e) {
-        this.communication.sendRendererShowErrorMessage(translations.processInstallationError)
-        throw new Error("Failed to install 'mysterium_client' process. " + e)
+        let messageForUser = translations.processInstallationError
+        if (e.message === 'User did not grant permission.') {
+          messageForUser = translations.processInstallationPermissionsError
+        }
+        this.communication.sendRendererShowErrorMessage(messageForUser)
+        throw new Error("Failed to install 'mysterium_client' process. " + e.message)
       }
     }
   }
--- a/src/app/mysterion.js
+++ b/src/app/mysterion.js
@@ -20,6 +20,7 @@
 import { app, BrowserWindow } from 'electron'
 import trayFactory from '../main/tray/factory'
 import { Installer, logLevels as processLogLevels } from '../libraries/mysterium-client'
+import { SUDO_PROMT_PERMISSION_DENIED } from '../libraries/mysterium-client/launch-daemon/installer'
 import translations from './messages'
 import MainMessageBusCommunication from './communication/main-message-bus-communication'
 import MainMessageBus from './communication/mainMessageBus'
@@ -235,7 +236,7 @@ class Mysterion {
         await this.installer.install()
       } catch (e) {
         let messageForUser = translations.processInstallationError
-        if (e.message === 'User did not grant permission.') {
+        if (e.message === SUDO_PROMT_PERMISSION_DENIED) {
           messageForUser = translations.processInstallationPermissionsError
         }
         this.communication.sendRendererShowErrorMessage(messageForUser)
--- a/src/app/mysterion.js
+++ b/src/app/mysterion.js
@@ -239,7 +239,7 @@ class Mysterion {
           messageForUser = translations.processInstallationPermissionsError
         }
         this.communication.sendRendererShowErrorMessage(messageForUser)
-        throw new Error("Failed to install 'mysterium_client' process. " + e.message)
+        throw new Error("Failed to install 'mysterium_client' process. " + e)
       }
     }
   }
--- a/src/app/mysterion.js
+++ b/src/app/mysterion.js
@@ -24,7 +24,7 @@ import { SUDO_PROMT_PERMISSION_DENIED } from '../libraries/mysterium-client/laun
 import translations from './messages'
 import MainMessageBusCommunication from './communication/main-message-bus-communication'
 import MainMessageBus from './communication/mainMessageBus'
-import {onFirstEvent, onFirstEventOrTimeout} from './communication/utils'
+import { onFirstEvent, onFirstEventOrTimeout } from './communication/utils'
 import path from 'path'
 import ConnectionStatusEnum from '../libraries/mysterium-tequilapi/dto/connection-status-enum'
 import logger from './logger'
@@ -143,7 +143,7 @@ class Mysterion {
     this.messageBus = new MainMessageBus(send, this.bugReporter.captureException)
     this.communication = new MainMessageBusCommunication(this.messageBus)
     this.communication.onCurrentIdentityChange((identityChange: CurrentIdentityChangeDTO) => {
-      const identity = new IdentityDTO({ id: identityChange.id })
+      const identity = new IdentityDTO({id: identityChange.id})
       this.bugReporter.setUser(identity)
     })
 
@@ -317,14 +317,20 @@ class Mysterion {
 
   _startProcess () {
     const cacheLogs = (level, data) => {
-      this.communication.sendMysteriumClientLog({ level, data })
+      this.communication.sendMysteriumClientLog({level, data})
       this.bugReporter.pushToLogCache(level, data)
     }
 
     logInfo("Starting 'mysterium_client' process")
     this.process.start()
-    this.process.onLog(processLogLevels.LOG, (data) => cacheLogs(processLogLevels.LOG, data))
-    this.process.onLog(processLogLevels.ERROR, (data) => cacheLogs(processLogLevels.ERROR, data))
+    try {
+      this.process.setupLogging()
+      this.process.onLog(processLogLevels.LOG, (data) => cacheLogs(processLogLevels.LOG, data))
+      this.process.onLog(processLogLevels.ERROR, (data) => cacheLogs(processLogLevels.ERROR, data))
+    } catch (e) {
+      logger.error('Failing to process logs. ', e)
+      this.bugReporter.captureException(e)
+    }
   }
 
   _startProcessMonitoring () {
--- a/src/libraries/create-file-if-missing.js
+++ b/src/libraries/create-file-if-missing.js
@@ -21,9 +21,18 @@ import fs from 'fs'
 import { promisify } from 'util'
 
 const writeFile = promisify(fs.writeFile)
+const openFile = promisify(fs.open)
+
+const READ_FILE_OR_FAIL_IF_NOT_EXIST = 'r'
 
 export default async function createFileIfMissing (path: string) {
-  if (!fs.existsSync(path)) {
-    await writeFile(path, '')
+  try {
+    await openFile(path, READ_FILE_OR_FAIL_IF_NOT_EXIST)
+  } catch (error) {
+    if (error.code && error.code === 'ENOENT') {
+      await writeFile(path, '')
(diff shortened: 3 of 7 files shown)


src/app/mysterion.js
```diff
@@ -234,8 +234,12 @@ class Mysterion {
       try {
         await this.installer.install()
       } catch (e) {
-        this.communication.sendRendererShowErrorMessage(translations.processInstallationError)
-        throw new Error("Failed to install 'mysterium_client' process. " + e)
+        let messageForUser = translations.processInstallationError
```
donce: `const`

src/app/mysterion.js
```diff
@@ -234,8 +234,12 @@ class Mysterion {
...
+        if (e.message === 'User did not grant permission.') {
```
donce: Depending on error message is dangerous - can we add some more distinct attribute? I.e.
```js
// throwing
const e = new Error('User did not grant permission.')
e.isPermissionError = true

//capturing
func isPermissionError(error: Error): Boolean {
  const errorObj = (error: Object)
  return errorObj.isPermissionError
}
```

src/app/messages.js
```diff
@@ -22,7 +22,8 @@ export default {
   initializationError: {
     message: 'Failed to load Mysterion'
   },
-  processInstallationError: 'Failed to install mysterium_client daemon. Please restart the app and grant permissions.',
+  processInstallationPermissionsError: 'Failed to install Mysterion. Please restart the app and grant permissions.',
```
donce: When user rejects terms, application closes instantly and this message is not seen. Are you sure we need this?

src/app/mysterion.js
donce: Or a `name` attribute: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name

src/app/mysterion.js
```diff
@@ -234,8 +234,12 @@ class Mysterion {
...
+          messageForUser = translations.processInstallationPermissionsError
+        }
+        this.communication.sendRendererShowErrorMessage(messageForUser)
+        throw new Error("Failed to install 'mysterium_client' process. " + e.message)
```
donce: We can just use `e`, not `e.message` as it was before - `e.toString()` adds error name as well, which we can customise.

src/renderer/App.vue
```diff
@@ -88,7 +88,7 @@
       })
 
       this.rendererCommunication.onShowRendererError((error) => {
-        logger.info('App error received from communication:', event)
+        logger.info('App error received from communication:', error)
```
donce: Wow, so all renderer errors were not working? :O
Can we add some test to ensure this does not happen again in the future?

src/libraries/mysterium-client/launch-daemon/installer.js
```diff
@@ -92,7 +96,20 @@ class Installer {
...
+    const errorFilePath = path.join(this.config.logDir, 'stderr.log')
+
+    if (!fs.existsSync(logFilePath)) {
+      await writeFile(logFilePath, null)
+    }
+
+    if (!fs.existsSync(errorFilePath)) {
+      await writeFile(errorFilePath, null)
+    }
+  }
```
donce: We can DRY part of method:
```js
async _createLogFilesIfMissing () {
  const logFilePath = path.join(this.config.logDir, 'stdout.log')
  await this._createFileIfMissing(logFilePath)
  const errorFilePath = path.join(this.config.logDir, 'stderr.log')
  await this._createFileIfMissing(errorFilePath)
}

async _createFileIfMissing (filePath) {
  if (!fs.existsSync(logFilePath)) {
    await writeFile(logFilePath, null)
  }
}
```

src/libraries/mysterium-client/launch-daemon/installer.js
```diff
@@ -92,7 +96,20 @@ class Installer {
     return !processInstalled() || this._pListChecksumMismatch()
   }
 
-  install () {
+  async _createLogFilesIfMissing () {
```
donce: Can we test this somehow?

src/app/mysterion.js
shroomist: sure, good idea. I'm filing MYS-83 task for this, its out of scope for now.

src/app/mysterion.js
shroomist: it's being changed on line 239

src/app/messages.js
shroomist: we need this. it's not terms, this is SUDO password dialog

src/app/mysterion.js
shroomist: ok

src/renderer/App.vue
shroomist: this check should be done with some compiler/pre-processor.
perhaps it's possible to setup jslint for this?

src/libraries/mysterium-client/launch-daemon/installer.js
shroomist: I know, just not that much of an improvement with added complexity.

src/libraries/mysterium-client/launch-daemon/installer.js
shroomist: done. please see `create-file-if-missing.spec.js`

test/unit/libraries/create-file-if-missing.spec.js
```diff
@@ -0,0 +1,53 @@
...
+  it('creates file if it doesn\'t exits', async () => {
+    await createFileIfMissing(createFilePath)
+    expect(existsSync(createFilePath)).to.be.true
+  })
+
+  it('doesn\'t overwrite file if it exists', async () => {
+    await createFileIfMissing(existingFilePath)
+    const existingFileContent = await readFilePromised(existingFilePath, {encoding: 'utf8'})
+    expect(existingFileContent).to.eql('some data')
+  })
```
donce: Beautiful 😍😍😍

src/libraries/mysterium-client/launch-daemon/installer.js
```diff
@@ -92,7 +97,12 @@ class Installer {
...
```
donce: Private methods should go after public methods.

src/libraries/create-file-if-missing.js
```diff
@@ -0,0 +1,29 @@
...
+// @flow
+
+import fs from 'fs'
+import { promisify } from 'util'
+
+const writeFile = promisify(fs.writeFile)
+
+export default async function createFileIfMissing (path: string) {
+  if (!fs.existsSync(path)) {
+    await writeFile(path, '')
```
donce: Why are we calling `existsSync` with **sync**, but `writeFile` with **async** - are there some reasons hidden? Because there is `.writeFileSync` method as well.

src/app/messages.js
donce: Oooh, ok!
Have you tried reproducing it, checking how this error message is shown - isn't it too long?

src/app/mysterion.js
```diff
@@ -234,7 +234,11 @@ class Mysterion {
       try {
         await this.installer.install()
       } catch (e) {
-        this.communication.sendRendererShowErrorMessage(translations.processInstallationError)
+        let messageForUser = translations.processInstallationError
+        if (e.message === 'User did not grant permission.') {
```
donce: Oh, so this message comes from [sudo-prompt](https://github.com/jorangreef/sudo-prompt/blob/ad291f7bd00bba09a01b7e4ce93dfa547f35f22d/index.js#L604). At first I though it was our error, but I was confused when I didn't find it in our code.

They doesn't export it, so we can't import that, but can we somehow clarify what this message is? I.e. introducing const with explaining name:

```js
const  SUDO_PROMT_PERMISSION_DENIED = 'User did not grant permission.'
```

src/libraries/mysterium-client/launch-daemon/installer.js
```diff
@@ -20,6 +20,11 @@ import sudo from 'sudo-prompt'
 import path from 'path'
 import md5 from 'md5'
 import { INVERSE_DOMAIN_PACKAGE_NAME, LAUNCH_DAEMON_PORT, PROPERTY_LIST_FILE, PROPERTY_LIST_NAME } from './config'
+import {promisify} from 'util'
+import createFileIfMissing from '../../create-file-if-missing'
+
+const writeFile = promisify(fs.writeFile)
+const sudoExec = promisify(sudo.exec)
 
 function processInstalled () {
```
ignasbernotas: why is this outside of the class?

shroomist: because it doesn't depend or change the class instance state

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -49,15 +51,23 @@ class Process {
       })
   }
 
-  onLog (level, cb) {
+  async onLog (level, cb) {
     if (level === logLevels.LOG) {
-      tailFile(path.join(this.logDirectory, 'stdout.log'), cb)
+      const filePath = path.join(this.logDirectory, 'stdout.log')
+      await createFileIfMissing(filePath)
```
donce: Invoking `process.onLog`, I wouldn't be expecting that it can create a new file - that is huge side-effect :O

Also, does each `onLog` starts-up a new file tail watcher? :O (that's out of PR scope, but this is solved by my suggestion as a bonus)

Can we somehow split this side-effect out of `onLog`? I.e.
```js
startLogging(level) {
  if (level == logLevels.LOG) {
    // ensure required files exist
    const filePath = path.join(this.logDirectory, 'stdout.log') 
    await createFileIfMissing(filePath)
    //tail all files
    tailFile(filePath, this._callback)
  }
  if (level === logLevels.ERROR) {
    // ...
  }
}

onLog(level, cb) {
  this._subscribers[level].push(cb)
}

_callback(level) {
  this._subscribers[level].forEach(cb => cb(...))
}
```

src/libraries/create-file-if-missing.js
```diff
@@ -0,0 +1,38 @@
...
+import { promisify } from 'util'
+
+const writeFile = promisify(fs.writeFile)
+const openFile = promisify(fs.open)
+
+const READ_FILE_OR_FAIL_IF_NOT_EXIST = 'r'
+
+export default async function createFileIfMissing (path: string) {
+  try {
+    await openFile(path, READ_FILE_OR_FAIL_IF_NOT_EXIST)
```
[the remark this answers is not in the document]
shroomist: this is the recommended way to do it.
I wanted to avoid using Sync methods when working with filesystem, and `fs.exists` (async) is deprecated.

src/app/messages.js
shroomist: screen shot 2018-06-12 at 11 48 58
its fine

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -31,6 +32,7 @@ class Process {
   /**
    * @constructor
    * @param {TequilapiClient} tequilapi - api to be used
+   * @param {string} daemonPort - port at which tequilapi is listening
```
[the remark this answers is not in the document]
shroomist: fixed

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -49,19 +58,26 @@ class Process {
...
-    throw new Error(`Unknown daemon logging level: ${level}`)
+  async setupLogging () {
+    await this._prepareLogFiles()
+    tailFile(this._stdoutPath, this._logCallback.bind(this, processLogLevels.LOG))
+    tailFile(this._stderrPath, this._logCallback.bind(this, processLogLevels.ERROR))
+    tailFile(SYSTEM_LOG, filterLine(INVERSE_DOMAIN_PACKAGE_NAME, this._logCallback.bind(this, processLogLevels.ERROR)))
+  }
+
+  onLog (level, cb) {
+    this._subscribers[level].push(cb)
```
donce: Currently, specific error is thrown for unknown level:
```js
throw new Error(`Unknown daemon logging level: ${level}`)
```
we should keep this - it's much better than getting undefined method error.

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -49,19 +58,26 @@ class Process {
...
-      tailFile(SYSTEM_LOG, filterLine(INVERSE_DOMAIN_PACKAGE_NAME, cb))
-      return
-    }
+  async _prepareLogFiles () {
+    await createFileIfMissing(this._stdoutPath)
+    await createFileIfMissing(this._stderrPath)
+  }
 
-    throw new Error(`Unknown daemon logging level: ${level}`)
+  async setupLogging () {
```
donce: Public method should go before private methods.

PR #195, opened 2018-06-08
Readded artefacts to builds

ignasbernotas:


.travis.yml
```diff
@@ -26,6 +26,7 @@ script:
 - util_scripts/check-headers.sh
 before_deploy:
 - bash util_scripts/travis-number.sh
+- FORCE_DOWNLOAD=1 yarn download:bins
```
tadovas: Without this line mysterion releases endup without mysterium client and other essential binaries. 👍

PR #199, opened 2018-06-11
Suppressed metric collection errors, added logging and improved code style

ignasbernotas:


src/app/statistics/aggregating-collector.js
```diff
@@ -15,31 +15,36 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import logger from '../logger'
```
shroomist: lets keep `// @flow` above imports

src/app/statistics/aggregating-collector.js
```diff
@@ -49,7 +54,12 @@ class AggregatingCollector implements EventCollector {
 
   async _sendEventsToDelegate (): Promise<any> {
     let eventsToSend = this._events.splice(0, this._accumulatorSize)
-    return this._delegate.collectEvents(...eventsToSend)
+
+    try {
+      await this._collector.collectEvents(...eventsToSend)
+    } catch (e) {
+      logger.error('Failed to send app metrics', e.stack)
```
shroomist: in this case we were getting `Error: Request failed with status code 503`
which seems more informative that `e.stack`

src/app/statistics/aggregating-collector.js
```diff
@@ -49,7 +54,12 @@ class AggregatingCollector implements EventCollector {
 
   async _sendEventsToDelegate (): Promise<any> {
```
donce: Since this method is not returning anything anymore (which is good!), it should be `Promise<void>`

src/app/statistics/aggregating-collector.js
donce: This can catch all sorts of problems, including programming problems - undefined method. If we're not capturing this error in Sentry, we will not know about this!

We should check if it's known error from tequilapi - than it's ok to ignore it, but in other cases, we can't ignore errors.

PR #200, opened 2018-06-11
Redirect to loading page when identity is not set

ignasbernotas:


test/unit/renderer/pages/Vpn.spec.js
```diff
@@ -0,0 +1,100 @@
...
+  async function mountComponent (identity = null, bugReporterCallback = null): Vue {
+    const localVue = createLocalVue()
+
+    const dependencies = new DIContainer(localVue)
+
+    dependencies.constant('bugReporter', {
+      setUser: () => {
+      },
+      captureInfoMessage: (message) => {
+        if (typeof bugReporterCallback === 'function') {
```
donce: Can we define type for `bugReporterCallback: ?((string) => void) = null` and check for `null` instead of type:
`if (bugReporterCallback != null)`?

Also, this object you're creating doesn't satisfy `BugReporter` interface - it's missing some of methods. Maybe it's worth having mock classes for `BugReporter`, as in `EmptyTequilapiClientMock` and `IdentityTequilapiClientMock`?

test/unit/renderer/pages/Vpn.spec.js
```diff
@@ -0,0 +1,100 @@
...
+          bugReporterCallback(message)
+        }
+      }
+    })
+
+    dependencies.constant('rendererCommunication', {
```
donce: This object doesn't satisfy `RendererCommunication` interface as well - we should have some mock implementation, as in `FakeMainCommunication`  - but it can be simpler, if we don't need recording messages in in `FakeMainCommunication`.

test/unit/renderer/pages/Vpn.spec.js
```diff
@@ -0,0 +1,100 @@
...
+        errors: errorStore
+      },
+      strict: false
+    })
+
+    const wrapper = mount(Vpn, {localVue, store, router})
+    return wrapper.vm
+  }
+
+  describe('Vpn Page', () => {
```
donce: `describe('Vpn Page')` inside `describe('Vpn')` - why do we need such complex nesting, when all we are describing is `Vpn` class?

src/renderer/pages/Vpn.vue
```diff
@@ -70,28 +72,41 @@
...
+      }
     },
     async mounted () {
       this.$store.dispatch(type.START_ACTION_LOOPING, new ActionLooperConfig(type.CONNECTION_IP, config.ipUpdateThreshold))
       this.$store.dispatch(type.START_ACTION_LOOPING, new ActionLooperConfig(type.FETCH_CONNECTION_STATUS, config.statusUpdateThreshold))
+
+      if (this.$store.getters.currentIdentity === null) {
+        this.bugReporter.captureInfoMessage('Redirecting from VPN to loader page')
+
+        this.$router.push('/load')
```
donce: Shouldn't this be in the beginning of `mounted`? Starting up action loopers when we exit from `Vpn` doesn't make much sense.

test/unit/renderer/pages/Vpn.spec.js
```diff
@@ -0,0 +1,100 @@
...
+import errorStore from '@/store/modules/errors'
+import Vpn from '@/pages/Vpn'
+
+import { nextTick } from '../../../helpers/utils'
+import { describe, it } from '../../../helpers/dependencies'
+
+import DIContainer from '../../../../src/app/di/vue-container'
+
+describe('Vpn', () => {
+  async function mountComponent (identity = null, bugReporterCallback = null): Vue {
```
donce: Types, please 😭

test/unit/renderer/pages/Vpn.spec.js
ignasbernotas: Why do we need full implementations when mocking? I prefer using the bare minimum that is actually used by the test subject. Implementing unused methods doesn't make much sense to me.

test/unit/renderer/pages/Vpn.spec.js
ignasbernotas: Woops.. :)

test/unit/renderer/pages/Vpn.spec.js
donce: In both approaches, there are advantages and disadvantages.

Personally, I like that Flow checks for me that I got method signatures right - and it helps keeping them right when refactoring.

But you have a good point, that mocking unused methods sucks.

To have advantages of both approaches, maybe something like with Tequilapi mocks would work - defining empty base mock class once and extending it with what you actually need, i.e.

```js
type EmptyMockBugReporter implements BugReporter {
  captureMessage(...) {
    throw new Error('This should not be invoked')
  }

  captureInfoMessage(...) {
    throw new Error('This should not be invoked')
  }

  ... other methods
}

type VpnBugReporterMock extends EmptyBugReporterMock {
  capturedMessage: ?String = null

  captureInfoMessage(message) {
    this.capturedMessage = message
  }
}
```

`EmptyMockBugReporter` class would be huge, but we would only need to implement it once.

As a bonus, we would get signatures and types checking, and also a clear error (Error('This should not be invoked')) when something is being invoked what we didn't expect to be invoked.

PR #202, opened 2018-06-12
Rename error capturing methods on BugReporter

donce:
This implicit *ERROR* level causes confusion, lets solve this.

### DIFF
--- a/src/app/bug-reporting/bug-reporter-main.js
+++ b/src/app/bug-reporting/bug-reporter-main.js
@@ -35,7 +35,7 @@ class BugReporterMain implements BugReporter {
     })
   }
 
-  captureMessage (message: string, context: ?any): void {
+  captureErrorMessage (message: string, context: ?any): void {
     this._captureMessage(message, 'error', context)
   }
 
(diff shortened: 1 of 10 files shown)


src/app/bug-reporting/bug-reporter-main.js
```diff
@@ -35,15 +35,15 @@ class BugReporterMain implements BugReporter {
...
+  captureErrorMessage (message: string, context: ?any): void {
     this._captureMessage(message, 'error', context)
   }
 
   captureInfoMessage (message: string, context: ?any): void {
     this._captureMessage(message, 'info', context)
   }
 
-  captureException (err: Error, context: ?any): void {
+  captureErrorException (err: Error, context: ?any): void {
```
shroomist: `Error Exception` sounds redundant to me

donce: Basically, these capturing methods differs by 2 parameters:
- **Object** which we capture - *message* or *exception*
- **Level** of captured event - *INFO* or *ERROR*

This is what's confusing - that *ERROR* level sounds too much like *exception* object :/

Yes, but I can't think of better names ATM :/

PR #205, opened 2018-06-12
Enable eslint-plugin-vue

donce:
I've enabled base rules for Vue - we will be able to adapt more rules gradually later on, by changing rules level: https://github.com/vuejs/eslint-plugin-vue#bulb-rules


donce:
> last part missing is flow checks in vue files

That's what I was actually trying to accomplish initially, but not much success (yet) :D

PR #209, opened 2018-06-14
Feature/mys 79 add new metric

mipo47:
Add new metrics to sentry tags:
identity_unlocked, proposals_fetched, connection_created, client_started, start_time
Additional data (extra):
connection_status, connection_statistics, connection_ip, last_health_check

Examples (logged 1, 10 and 20 seconds after Mysterion started): 
https://sentry.io/mysterium-network/mysterion/issues/579291150/?query=is:unresolved
https://sentry.io/mysterium-network/mysterion/issues/579291366/?query=is:unresolved
https://sentry.io/mysterium-network/mysterion/issues/579291731/?query=is:unresolved


src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,130 @@
...
+  for (let i = 0; i < aProps.length; i++) {
+    const propName = aProps[i]
+    if (a[propName] !== b[propName]) {
+      return false
+    }
+  }
+  return true
+}
+
+class BugReporterMetrics {
```
donce: Instead of global state, we can use DI.

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,130 @@
...
+  // Create arrays of property names
+  const aProps = Object.getOwnPropertyNames(a)
+  const bProps = Object.getOwnPropertyNames(b)
+
+  if (aProps.length !== bProps.length) {
+    return false
+  }
+
+  for (let i = 0; i < aProps.length; i++) {
+    const propName = aProps[i]
```
donce: We can iterate in a functional style, something like this:
```js
if (aProps.some((prop) => a[propName !== b[propName])) {
  return false
}
```

src/app/bug-reporting/bug-reporter-metrics.js
donce: Could you add a short description for class as a comment?

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,130 @@
...
+    })
+    this._extra.forEach((value, metric) => {
+      data.extra[metric] = value
+    })
+  }
+
+  dateTimeString (): string {
+    return (new Date()).toUTCString()
+  }
+}
```
donce: No tests :(

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,130 @@
...
+
+    metricStorage.set(metric, value)
+
+    if (this._messageBus) {
+      const data: MetricSyncDTO = {
+        metric: metric,
+        value: value
+      }
+      this._messageBus.send(messages.METRIC_SYNC, data)
+    }
```
donce: To avoid unnecessary nesting, we can use conditional guards in such cases: https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

i.e.
```js
if (!this._messageBus) {
  return
}

const data: MetricSyncDTO = {
...
```

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,130 @@
...
+    this.set(this.StartTime, UNKNOWN, this._tags)
+    this.set(this.ConnectionStatus, UNKNOWN, this._extra)
+    this.set(this.ConnectionStatistics, UNKNOWN, this._extra)
+    this.set(this.ConnectionIP, UNKNOWN, this._extra)
+    this.set(this.HealthCheckTime, UNKNOWN, this._extra)
+  }
+
+  syncWith (messageBus: MessageBus) {
+    this._messageBus = messageBus
+    this._messageBus.on(messages.METRIC_SYNC, (data: any) => {
```
donce: Message bus is just a transfer implementation - Communication is abstraction on top of that, which provides clear typed interfaces.

For consistency, please add `onMetricSync` to `MainCommunication` and `sendMetricSync` to `RendererCommunication` and use it here instead of raw message bus.

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,130 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+// @flow
+
+import type {MessageBus} from '../communication/messageBus'
+import messages from '../communication/messages'
+import type {MetricSyncDTO} from '../communication/dto'
+
+const UNKNOWN = 'no'
```
donce: Maybe something like `<unknown>` would be easier to understand when inspecting issues in Sentry?

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,130 @@
...
+
+  syncWith (messageBus: MessageBus) {
+    this._messageBus = messageBus
+    this._messageBus.on(messages.METRIC_SYNC, (data: any) => {
+      const dto: MetricSyncDTO = (data: MetricSyncDTO)
+      this.set(dto.metric, dto.value)
+    })
+  }
+
+  set (metric: string, value: mixed, metricStorage: ?Map<string, any> = null) {
```
donce: For `metric` param, maybe it's possible somehow to specify list of accepted values? We might use unions - they are similar to enums: https://flow.org/en/docs/types/unions/

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    metricStorage.set(metric, value)
+
+    if (this._messageBus) {
+      const data: MetricSyncDTO = {
+        metric: metric,
+        value: value
+      }
+      this._messageBus.send(messages.METRIC_SYNC, data)
```
donce: So metrics, set before invoking `syncWith`, won't be visible in renderer?

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,130 @@
...
+    // set initial values for metrics
+    this.set(this.IdentityUnlocked, UNKNOWN, this._tags)
+    this.set(this.ProposalsFetched, UNKNOWN, this._tags)
+    this.set(this.ConnectionCreated, UNKNOWN, this._tags)
+    this.set(this.ClientStarted, UNKNOWN, this._tags)
+    this.set(this.StartTime, UNKNOWN, this._tags)
+    this.set(this.ConnectionStatus, UNKNOWN, this._extra)
+    this.set(this.ConnectionStatistics, UNKNOWN, this._extra)
+    this.set(this.ConnectionIP, UNKNOWN, this._extra)
+    this.set(this.HealthCheckTime, UNKNOWN, this._extra)
```
donce: So it was surprise that these `set` calls have 2 purposes:
- set default values `UNKNOWN`
- map each metric to a specific storage (tags/extra)

Can we make mapping of each metric to storage more obvious, i.e. have that in a `Map`? That way, we will be able to get rid of complex mapping resolution in `set`:

```js
    if (!metricStorage) {
      if (this._tags.has(metric)) {
        this.set(metric, value, this._tags)
      } else {
        this.set(metric, value, this._extra)
      }
      return
    }
```

src/libraries/mysterium-tequilapi/client.js
```diff
@@ -101,12 +105,16 @@ class HttpTequilapiClient implements TequilapiClient {
     const responseDto = new ProposalsResponseDTO(response)
 
     if (!responseDto.proposals) {
+      bugReporterMetrics.set(bugReporterMetrics.ProposalsFetched, false)
       return []
     }
+    bugReporterMetrics.set(bugReporterMetrics.ProposalsFetched, true)
+    // $FlowFixMe
```
donce: Solution for this is in #frontend channel in Slack.

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,130 @@
...
+  const aProps = Object.getOwnPropertyNames(a)
+  const bProps = Object.getOwnPropertyNames(b)
+
+  if (aProps.length !== bProps.length) {
+    return false
+  }
+
+  for (let i = 0; i < aProps.length; i++) {
+    const propName = aProps[i]
+    if (a[propName] !== b[propName]) {
```
donce: What if prop is an `Object` itself? Shouldn't we invoke `isEquivalent(a[propName], b[propName])` ?
If that's true, can you make a test to capture this case?

src/app/bug-reporting/bug-reporter-metrics.js
mipo47: now added

src/app/bug-reporting/bug-reporter-metrics.js
mipo47: for this case it's ok :)

src/app/bug-reporting/bug-reporter-metrics.js
mipo47: yes

src/app/bug-reporting/bug-reporter-metrics.js
mipo47: It was in my first version, but 'no' looks better (less text, easier to understand)

Waldz: Guys, it's `null`.
Is there a reason for this over-engineering a `null`?

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,126 @@
...
+type MetricTags = $Values<typeof TAGS>
+type MetricExtra = $Values<typeof EXTRA>
+type Metric = MetricTags | MetricExtra
+
+/**
+ * Collects and synchronizes data used in BugReporter
+ */
+class BugReporterMetrics {
+  Tags = TAGS
+  Extra = EXTRA
```
Waldz: No point to copy values, it's constants

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,126 @@
...
+      const dto: MetricSyncDTO = (data: MetricSyncDTO)
+      this.set(dto.metric, dto.value)
+    })
+  }
+
+  set (metric: Metric, value: mixed, metricStorage: ?Map<Metric, any> = null): void {
+    // detect default storage for metric (tags or extra)
+    if (!metricStorage) {
+      if (this._tags.has(metric)) {
+        this.set(metric, value, this._tags)
```
Waldz: Why recursion needed? For me overcomplicates understanding a function

src/renderer/dependencies/modules/application.js
```diff
@@ -37,7 +38,9 @@ function bootstrap (container: Container) {
     'rendererCommunication',
     [],
     () => {
-      return new RendererCommunication(new RendererMessageBus())
+      const messageBus = new RendererMessageBus()
```
Waldz: Reuse existing instance of `messageBus` via DI container

src/libraries/mysterium-tequilapi/client.js
```diff
@@ -29,6 +29,7 @@ import ConnectionStatusDTO from './dto/connection-status'
 import ConnectionRequestDTO from './dto/connection-request'
 import ConsumerLocationDTO from './dto/consumer-location'
 import {TIMEOUT_DISABLED} from './timeouts'
+import {bugReporterMetrics} from '../../app/bug-reporting/bug-reporter-metrics'
```
Waldz: We are seeking to move things in `libraries/...` out of project as packages.
Dont put application specific logic here

src/libraries/mysterium-tequilapi/client.js
```diff
@@ -89,7 +91,9 @@ class HttpTequilapiClient implements TequilapiClient {
   }
 
   async identityUnlock (id: string, passphrase: string): Promise<void> {
+    bugReporterMetrics.set(bugReporterMetrics.Tags.IdentityUnlocked, false)
```
Waldz: Wrong place for marking identity as unlocked. It should be then mysteriumClient crashes: `this.monitoring.subscribeDown(...)`

src/libraries/mysterium-tequilapi/client.js
```diff
@@ -99,14 +103,18 @@ class HttpTequilapiClient implements TequilapiClient {
...
+      bugReporterMetrics.set(bugReporterMetrics.Tags.ProposalsFetched, false)
       return []
     }
-    return responseDto.proposals
+    bugReporterMetrics.set(bugReporterMetrics.Tags.ProposalsFetched, true)
+    return proposals
   }
 
   async connectionCreate (request: ConnectionRequestDTO, timeout: ?number = TIMEOUT_DISABLED): Promise<ConnectionStatusDTO> {
+    bugReporterMetrics.set(bugReporterMetrics.Tags.ConnectionCreated, false)
```
Waldz: Wrong place to understand state of connection, currently connection is owned by `src/renderer/store/modules/connection.js`

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,126 @@
...
+const TAGS = {
+  IdentityUnlocked: 'identity_unlocked',
+  ProposalsFetched: 'proposals_fetched',
+  ConnectionCreated: 'connection_created',
+  ClientStarted: 'client_started',
+  StartTime: 'start_time'
+}
+
+const EXTRA = {
+  HealthCheckTime: 'last_health_check',
```
Waldz: I see no point of separating constants into 2 groups, just duplicates a lot of same code

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,126 @@
...
+  get (metric: Metric): any {
+    if (this._tags.has(metric)) {
+      return this._tags.get(metric)
+    } else if (this._extra.has(metric)) {
+      return this._extra.get(metric)
+    }
+    return null
+  }
+
+  addMetricsTo (data: any): void {
```
Waldz: You dont even know that type of object it is, and append to any type.
Also could be replaced just with `getAll(): Map<MetricTags, any>`

src/app/bug-reporting/bug-reporter-metrics.js
Waldz: `_messageBus` looks like a required dependency. Otherwise we will have side effect of some un-synced metrics.

Our approach: 
- require dependency in constructor
- BugReporterMetrics is booted with all application start to DI container
- avoid booting via global scope `import '../bug-reporter-metrics'`, instead load from CI container

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,60 @@
...
+  it('sets extra metric', () => {
+    const metricKey = bugReporterMetrics.Extra.HealthCheckTime
+    const metricValue = bugReporterMetrics.dateTimeString()
+
+    expect(bugReporterMetrics.get(metricKey)).to.eql(UNKNOWN)
+    bugReporterMetrics.set(metricKey, metricValue)
+    expect(bugReporterMetrics.get(metricKey)).to.eql(metricValue)
+  })
+
+  it('add metrics to object', () => {
```
Waldz: Missing tests for sync logic

src/app/bug-reporting/bug-reporter-metrics.js
mipo47: it's shortcuts to constants, no need to import TAGS, EXTRA to set metrics

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,126 @@
...
+    }
+    for (let tagKey of Object.keys(this.Extra)) {
+      this.set(this.Extra[tagKey], UNKNOWN, this._extra)
+    }
+  }
+
+  syncWith (messageBus: MessageBus): void {
+    this._messageBus = messageBus
+    this._messageBus.on(messages.METRIC_SYNC, (data: any) => {
+      const dto: MetricSyncDTO = (data: MetricSyncDTO)
```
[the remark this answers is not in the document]
mipo47: Flow doesn't allows to cast mixed type to MetricSyncDTO, but allow to cast mixed to any and any to MetricSyncDTO

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
mipo47: Added

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,108 @@
...
+
+  get (metric: Metric): any {
+    return this._metrics.get(metric)
+  }
+
+  _setValues (metrics: Array<Metric>, dest: any) {
+    for (let metric of metrics) {
+      dest[metric] = this.get(metric) || 'no'
+    }
+  }
```
donce: Public methods should go before private methods.

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,108 @@
...
+
+  addMetricsTo (data: RavenData): void {
+    this._setValues((Object.values(TAGS): any), data.tags)
+    this._setValues((Object.values(EXTRA): any), data.extra)
+  }
+
+  dateTimeString (): string {
+    return (new Date()).toUTCString()
+  }
```
donce: This doesn't seem to belong to this class - this method doesn't need class and class doesn't need this method. Maybe it's better to keep this outside of class?

src/app/mysterion.js
```diff
@@ -90,6 +91,8 @@ class Mysterion {
   }
 
   run () {
+    bugReporterMetrics.set(METRICS.StartTime, bugReporterMetrics.dateTimeString())
```
donce: So this metric will only be visible when capture happens in Main process? It's fine for me at this point, but can we add some TODO to make sure this get to renderer in the future?

src/libraries/mysterium-tequilapi/client.js
```diff
@@ -89,7 +91,9 @@ class HttpTequilapiClient implements TequilapiClient {
   }
 
   async identityUnlock (id: string, passphrase: string): Promise<void> {
+    bugReporterMetrics.set(METRICS.IdentityUnlocked, false)
```
donce: hmm, so `IdentityUnlocked` metric will have 3 states:
- `'no'` initially
- `false` just before starting unlock
- `true` if unlocking succeeds

Maybe we can clarify this somehow - i.e.
- `'no'` (or even `'not started'`, but this might be tricky to implement)
- `'started'`
- `'unlocked'`

src/libraries/mysterium-tequilapi/client.js
```diff
@@ -99,14 +103,18 @@ class HttpTequilapiClient implements TequilapiClient {
       throw new Error('Proposals response body is missing')
     }
     const responseDto = new ProposalsResponseDTO(response)
+    const proposals = responseDto.proposals
 
-    if (!responseDto.proposals) {
+    if (!proposals) {
+      bugReporterMetrics.set(METRICS.ProposalsFetched, false)
```
donce: `const response = await this.http.get('proposals', query)` can raise an exception - this that case this line will not be executed and `ProposalsFetched` will be left unset. We should set it to `false` on all failures.

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/**
+ * Collects and synchronizes data used in BugReporter
+ */
+class BugReporterMetrics {
+  _metrics: Map<Metric, any> = new Map()
+  _messageBus: ?MessageBus = null
+
+  syncWith (messageBus: MessageBus): void {
```
donce: Please use `MainCommunication`, instead of `MessageBus` - the later is just a transport implementation.

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,108 @@
...
+      this._messageBus.send(messages.METRIC_SYNC, data)
+    }
+  }
+
+  get (metric: Metric): any {
+    return this._metrics.get(metric)
+  }
+
+  _setValues (metrics: Array<Metric>, dest: any) {
+    for (let metric of metrics) {
```
donce: `const`

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+// @flow
+import {describe, it, expect} from '../../../helpers/dependencies'
+import {bugReporterMetrics, EXTRA, METRICS, TAGS} from '../../../../src/app/bug-reporting/bug-reporter-metrics'
+import FakeMessageBus from '../../../helpers/fakeMessageBus'
+import type {MetricSyncDTO} from '../../../../src/app/communication/dto'
+
+describe('BugReporterMetrics', () => {
+  it('sets tag metric', () => {
```
donce: Specs should be grouped into method describtions.

src/libraries/mysterium-tequilapi/client.js
```diff
@@ -127,20 +135,21 @@ class HttpTequilapiClient implements TequilapiClient {
     if (!response) {
       throw new Error('Connection status response body is missing')
     }
-
+    bugReporterMetrics.set(METRICS.ConnectionStatus, response)
```
donce: What get serialised here if we pass class instance to Sentry - `Object.toString()`? Isn't `[object Object]` what gets serialized?

src/renderer/dependencies/modules/application.js
```diff
@@ -37,7 +38,9 @@ function bootstrap (container: Container) {
...
+      bugReporterMetrics.syncWith(messageBus)
```
donce: It's a bit unexpected that just by using some component from DI, we are actually making important side-effects. We should put this side-effect somewhere explicitly, not inside building of DI.

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
donce: I.e.

```js
describe('BugReporterMetrics', () => {
  describe('.set', () => {
    it('sets extra metric', () => ...)
  })
})
```

That way, we get a clear hierarchy and `it('sets extra metric'` is specifying behaviour of an object - `.set` method.

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,90 @@
...
+    const metricKey = METRICS.IdentityUnlocked
+    const metricValue = true
+
+    expect(bugReporterMetrics.get(metricKey)).to.be.eql(undefined)
```
donce: `.to.be.undefined`

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,108 @@
...
+    if (this._messageBus) {
+      const data: MetricSyncDTO = {
+        metric: metric,
+        value: value
+      }
+      this._messageBus.send(messages.METRIC_SYNC, data)
+    }
+  }
+
+  get (metric: Metric): any {
```
donce: Is this method exposed just for tests? If so, can't we encapsulate this and expose only `addMetricsTo`? We can test using `addMetricsTo` method - than we would be testing what we'll actually use.

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,90 @@
...
+  it('sets extra metric', () => {
+    const metricKey = METRICS.HealthCheckTime
+    const metricValue = bugReporterMetrics.dateTimeString()
+
+    expect(bugReporterMetrics.get(metricKey)).to.eql(undefined)
+    bugReporterMetrics.set(metricKey, metricValue)
+    expect(bugReporterMetrics.get(metricKey)).to.eql(metricValue)
+  })
+
+  it('add metrics to object', () => {
```
donce: it add**s**

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,90 @@
...
+    bugReporterMetrics.addMetricsTo(data)
+    const tagKeys = Object.keys(data.tags)
+    const extraKeys = Object.keys(data.extra)
+
+    expect(tagKeys.length + extraKeys.length).to.eql(Object.values(METRICS).length)
+
+    for (let tagKey of Object.values(TAGS)) {
+      expect(tagKeys).contains(tagKey)
+    }
+    for (let extraKey of Object.values(EXTRA)) {
```
donce: `const`

src/app/bug-reporting/bug-reporter-metrics.js
ignasbernotas: I suggest to use 'N/A' instead of 'no'

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,90 @@
...
+      extra: {}
+    }
+    bugReporterMetrics.addMetricsTo(data)
+    const tagKeys = Object.keys(data.tags)
+    const extraKeys = Object.keys(data.extra)
+
+    expect(tagKeys.length + extraKeys.length).to.eql(Object.values(METRICS).length)
+
+    for (let tagKey of Object.values(TAGS)) {
+      expect(tagKeys).contains(tagKey)
```
donce: **expect tagKeys contains** doesn't sound very english - usually `expect` has `.to` afterwards:
```expect(tagKeys).to.contain(tagKey)` - that way it would be read as **expect tagKeys to contain tagKey**

File: test/unit/app/bug-reporting/bug-reporter-metrics.spec.js [hunk 37]
```diff
@@ -0,0 +1,90 @@
...
+    }
+    for (let extraKey of Object.values(EXTRA)) {
+      expect(extraKeys).contains(extraKey)
+    }
```
donce: We are not actually testing that `addMetricsTo` sets **correct** values - we just check that it sets some values.

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,90 @@
...
+  })
+
+  it('send/receive metric via message bus', () => {
```
donce: *it sends and receives metrics via message bus*?

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,90 @@
...
+    const messageBus = new FakeMessageBus()
+    bugReporterMetrics.syncWith(messageBus)
+
+    const metricKey = METRICS.ConnectionIP
+    const metricValue = '{ip: "127.0.0.1"}'
+    bugReporterMetrics.set(metricKey, metricValue)
+
+    expect(messageBus.lastData).to.not.be.eql(null)
```
donce: `.to.not.be.null`. Also, same in other places.

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,90 @@
...
+    const dto: MetricSyncDTO = (messageBus.lastData: any)
+    expect(dto).to.not.be.eql(null)
```
donce: This assert doesn't do much actually - there is no runtime typecasting in Flow, so we don't need to test that.

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,90 @@
...
+      const updateDto: MetricSyncDTO = {
+        metric: metricKey,
+        value: newValue
+      }
+      messageBus.triggerOn(messageBus.lastChannel, updateDto)
+
+      const updatedValue = bugReporterMetrics.get(metricKey)
+      expect(updatedValue).to.be.eql(newValue)
+    } else {
+      expect(messageBus.lastChannel).to.not.be.eql(null)
```
donce: I don't quite get why do we need `if` inside a test - and and why `else` just expects what we already know to be false?
Can't we just `expect(messageBus.lastChannel).not.to.be.null`?

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,115 @@
...
+ */
+// @flow
+
+import type {MessageBus} from '../communication/messageBus'
+import messages from '../communication/messages'
+import type {MetricSyncDTO} from '../communication/dto'
+
+/**
+ * Used as default metric value in Sentry
+ * Later metric value will be replaces using method 'set'
```
donce: `Later metric value will be replaces` - can't understand this.

src/app/bug-reporting/bug-reporter-metrics.js
donce: Please check this comment.

src/app/bug-reporting/bug-reporter-metrics.js
donce: As discussed, we might try adding new interface, i.e. `MetricsCommunication`, which could be implemented by both `MainMessageBusCommunication` and `RendererMessageBusCommunication`

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,115 @@
...
+  set (metric: Metric, value: mixed): void {
+    const oldValue = this._metrics.get(metric)
+    if (JSON.stringify(oldValue) === JSON.stringify(value)) {
+      // metric's value was not updated
+      return
+    }
+
+    this._metrics.set(metric, value)
+
+    if (this._messageBus) {
```
donce: Please add `TODO` to indicate that metrics set before `syncWith` should be sent to communication as well.

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,93 @@
...
+import {bugReporterMetrics, EXTRA, METRICS, NOT_SET, TAGS} from '../../../../src/app/bug-reporting/bug-reporter-metrics'
+import FakeMessageBus from '../../../helpers/fakeMessageBus'
+import type {MetricSyncDTO} from '../../../../src/app/communication/dto'
+
+describe('BugReporterMetrics', () => {
+  it('sets tag metric', () => {
+    const metricKey = METRICS.IdentityUnlocked
+    const metricValue = true
+
+    expect(bugReporterMetrics.get(metricKey)).to.be.undefined
```
donce: Please test `BugReporterMetrics` class directly instead of singleton to avoid having side-effects in singleton between tests.

src/libraries/mysterium-tequilapi/client.js
```diff
@@ -62,6 +63,7 @@ class HttpTequilapiClient implements TequilapiClient {
       throw new Error('Healthcheck response body is missing')
     }
 
+    bugReporterMetrics.set(METRICS.HealthCheckTime, bugReporterMetrics.dateTimeString())
```
donce: These should not be inside `TequilapiClient`, as discussed.

src/libraries/mysterium-tequilapi/client.js
donce: Please check this comment.

src/renderer/dependencies/modules/application.js
donce: Please check this.

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
donce: Please check this.

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Collects and synchronizes data used in BugReporter
+ */
+export class BugReporterMetrics extends MapSync<Metric> {
+  dateTimeString (): string {
+    return (new Date()).toUTCString()
+  }
+
+  addMetricsTo (data: RavenData): void {
+    this._setValues((Object.values(TAGS): any), data.tags)
```
shroomist: interesting syntax here, describing return type for function that is being invoked?

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Used as default metric value in Sentry
+ * Later metric value will be replaces using method 'set'
+ * @type {string}
+ */
+const NOT_SET = 'N/A'
+
+const TAGS = {
+  IdentityUnlocked: 'identity_unlocked',
```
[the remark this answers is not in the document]
mipo47: Object will be converted to type

src/libraries/tequilapi-metrics.js
```diff
@@ -0,0 +1,111 @@
...
+  constructor (client: TequilapiClient, bugReporterMetrics: BugReporterMetrics) {
+    this.client = client
+    this.bugReporterMetrics = bugReporterMetrics
+  }
+
+  async stop (): Promise<void> {
+    return this.client.stop()
+  }
+
+  async identitiesList (): Promise<Array<IdentityDTO>> {
```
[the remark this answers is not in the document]
mipo47: super. doesn't works in karma tests

src/libraries/mysterium-tequilapi/client.js
```diff
@@ -99,11 +96,12 @@ class HttpTequilapiClient implements TequilapiClient {
       throw new Error('Proposals response body is missing')
     }
     const responseDto = new ProposalsResponseDTO(response)
+    const proposals = responseDto.proposals
```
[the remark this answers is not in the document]
mipo47: Tried to keep same style in file: empty line count before return

src/libraries/tequilapi-metrics.js
```diff
@@ -0,0 +1,111 @@
...
+
+  async identityUnlock (id: string, passphrase: string): Promise<void> {
+    this.bugReporterMetrics.set(METRICS.IDENTITY_UNLOCKED, false)
+    await this.client.identityUnlock(id, passphrase)
+    this.bugReporterMetrics.set(METRICS.IDENTITY_UNLOCKED, true)
+  }
+
+  async findProposals (filter: ?ProposalsFilter): Promise<Array<ProposalDTO>> {
+    const result = await this.client.findProposals(filter)
+    if (!result || result.length === 0) {
```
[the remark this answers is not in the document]
mipo47: fixed

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,106 @@
...
+import FakeMainCommunication from '../../../helpers/fakeMainCommunication'
+
+describe('BugReporterMetrics', () => {
+  let bugReporterMetrics : BugReporterMetrics
+
+  beforeEach(() => {
+    bugReporterMetrics = new BugReporterMetrics()
+  })
+
+  describe('get/set', () => {
```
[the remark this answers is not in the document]
mipo47: Added new test groups

src/dependencies/bug-reporting.js
```diff
@@ -54,6 +61,9 @@ function bootstrap (container: Container) {
           platform_release: os.release()
         },
         dataCallback: (data) => {
+          const metrics = bugReporterMetrics.getMetrics()
+          Object.assign(data.tags, metrics.tags)
+          Object.assign(data.extra, metrics.extra)
```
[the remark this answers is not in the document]
mipo47: addMetrics is thrown away now

test/unit/libraries/mysterium-tequilapi/client.spec.js
```diff
@@ -35,7 +35,7 @@ describe('HttpTequilapiClient', () => {
   let mock
   beforeEach(() => {
     const axioInstance = axios.create()
-    api = new HttpTequilapiClient(new AxiosAdapter(axioInstance), 1)
```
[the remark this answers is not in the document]
mipo47: Thanks

src/dependencies/mysterium-tequilapi.js
```diff
@@ -28,9 +34,18 @@ function bootstrap (container: Container) {
...
+    ['bugReporterMetrics', 'tequilapiClient.config'],
+    (bugReporterMetrics: BugReporterMetrics, config: Object) => {
+      const axiosInstance = axios.create({
+        baseURL: config.baseURL,
+        headers: {
+          'Cache-Control': 'no-cache, no-store'
+        }
+      })
+      const axiosAdapter = new AxiosAdapter(axiosInstance, TIMEOUT_DEFAULT)
+      const client: TequilapiClient = new HttpTequilapiClient(axiosAdapter)
```
Waldz: Tequila factoring still can be DRY `client = tequilapiClientFactory(config.baseURL)`

src/libraries/mysterium-tequilapi/client-factory.js
```diff
@@ -19,20 +19,21 @@
...
-import HttpTequilapiClient from './client'
 import {TIMEOUT_DEFAULT} from './timeouts'
+import {BugReporterMetrics} from '../../app/bug-reporting/bug-reporter-metrics'
+import HttpTequilapiClient from './client'
+import type {TequilapiClient} from './client'
 
 const TEQUILAPI_URL = 'http://127.0.0.1:4050'
 
-function tequilapiClientFactory (baseUrl: string = TEQUILAPI_URL, defaultTimeout: number = TIMEOUT_DEFAULT) {
+function tequilapiClientFactory (bugReporterMetrics: ?BugReporterMetrics = null, baseUrl: string = TEQUILAPI_URL, defaultTimeout: number = TIMEOUT_DEFAULT): TequilapiClient {
```
Waldz: Why?

test/integration/libraries/mysterium-client/standalone/process.spec.js
```diff
@@ -21,18 +21,27 @@ import sleep from '../../../../../src/libraries/sleep'
...
 
 xdescribe('Standalone Process', () => {
-  let process, tequilapi
+  let process
   const logs = []
   const tequilapiPort = 4055
   const clientBinDirectory = path.resolve(__dirname, '../../../../../bin')
   const tmpDirectory = os.tmpdir()
 
+  const axiosInstance = axios.create({
```
Waldz: No point to change here

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,109 @@
...
+    })
+  })
+
+  describe('getMetrics', () => {
+    it('gets all metrics', () => {
+      const data = bugReporterMetrics.getMetrics()
+      const tagKeys = Object.keys(data.tags)
+      const extraKeys = Object.keys(data.extra)
+
+      expect(tagKeys.length + extraKeys.length).to.eql(Object.values(METRICS).length)
```
Waldz: You dont need to test JS + operator

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,109 @@
...
+
+      for (let tagKey of Object.values(TAGS)) {
+        expect(tagKeys).to.contain(tagKey)
+        // $FlowFixMe
+        expect(data.tags[tagKey]).to.be.eql(mapSync.get(tagKey) || NOT_SET)
```
Waldz: You asserting with conditional, not point of such test.

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,109 @@
...
+
+  describe('getMetrics', () => {
+    it('gets all metrics', () => {
+      const data = bugReporterMetrics.getMetrics()
+      const tagKeys = Object.keys(data.tags)
+      const extraKeys = Object.keys(data.extra)
+
+      expect(tagKeys.length + extraKeys.length).to.eql(Object.values(METRICS).length)
+
+      for (let tagKey of Object.values(TAGS)) {
```
Waldz: No point to do complicated logic tests.
I recommend simply calling several setters, and expect result of `getMetricts()`

Waldz: More static tests are, more stable it is

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,109 @@
...
+      const communication = new FakeMapSyncCommunication()
+      bugReporterMetrics.syncWith(communication)
+
+      let lastUpdate: ?MapSyncDTO<Metric> = null
+      communication.onMapUpdate(update => {
+        lastUpdate = update
+      })
+
+      const metricKey = METRICS.CONNECTION_IP
+      const metricValue = '{ip: "127.0.0.1"}'
```
Waldz: Strange to see unparsed data, for reasons expect to have just values in this storage.
```
CONNECTION_IP = '127.0.0.1'
CONNECTION_STATUS = 'connected'
CONNECTION_STATISTICS = {
  bytes_sent: 3564574, 
  bytes_received: 134264
}
```

Reasons:
- Sentry size limitation
- Storage allows indexing, searching
- Usually logs with verbosity are for detailed data (request/response/payload),

src/app/bug-reporting/bug-reporter-metrics.js
```diff
@@ -0,0 +1,103 @@
...
+  extra: keyValueMap,
+}
+
+/**
+ * Collects and synchronizes data used in BugReporter
+ */
+class BugReporterMetrics {
+  _mapSync: MapSync<Metric>
+
+  constructor (mapSync: ?MapSync<Metric> = null) {
```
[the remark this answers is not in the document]
mipo47: Ok, fixed

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
mipo47: why?

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
mipo47: It simplifies implementation, no need to know tequilapi format and do data mapping

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
mipo47: Test doesn't know about Metrics we use. Tried to made universal test for any metrics

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
mipo47: This is test for missed metrics and also + operator :)

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,135 @@
...
+    it('sets tag metric', () => {
+      const metricKey = METRICS.IDENTITY_UNLOCKED
+      const metricValue = true
+
+      expect(mapSync.get(metricKey)).to.be.undefined
+      bugReporterMetrics.set(metricKey, metricValue)
+      expect(mapSync.get(metricKey)).to.eql(metricValue)
+    })
+
+    it('sets extra metric', () => {
```
Waldz: 2 absolutely identical tests

src/renderer/App.vue
```diff
@@ -69,6 +69,8 @@ export default {
     }
   },
   async mounted () {
+    this.bugReporterMetrics.syncWith(this.rendererCommunication)
```
Waldz: Why here not  2 lines down?
I mean, lets not build code which is meant to have race conditions.

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
donce: Agree - these two tests doesn't differ, because `MapSync` doesn't differ `EXTRA` from `TAGS`.

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,135 @@
...
+
+      for (let tagKey of Object.values(TAGS)) {
+        expect(tagKeys).to.contain(tagKey)
+        // $FlowFixMe
+        expect(data.tags[tagKey]).to.deep.equal(mapSync.get(tagKey) || NOT_SET)
+      }
+      for (let extraKey of Object.values(EXTRA)) {
+        expect(extraKeys).to.contain(extraKey)
+        // $FlowFixMe
+        expect(data.extra[extraKey]).to.deep.equal(mapSync.get(extraKey) || NOT_SET)
```
donce: Maybe it's enough to test that all keys are present? Values are already (somewhat) covered by other tests.

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,135 @@
...
+      }
+
+      expect(data).to.deep.equal(bugReporterMetrics.getMetrics())
```
donce: What's the purpose of this assertion?

test/unit/app/bug-reporting/bug-reporter-metrics.spec.js
```diff
@@ -0,0 +1,135 @@
...
+    })
+  })
+
+  describe('syncWith', () => {
+    it('sends/receives metric via message bus', () => {
```
donce: Looks like to tests: `it sends metric via message bus` and `it receives metric via message bus`.

test/unit/app/bug-reporting/tequilapi-metrics.spec.js
```diff
@@ -0,0 +1,48 @@
...
+
+  beforeEach(() => {
+    api = new EmptyTequilapiClientMock()
+    mapSync = new MapSync()
+    metrics = new BugReporterMetrics(mapSync)
+    apiMetrics = new TequilapiClientWithMetrics(api, metrics)
+  })
+
+  describe('healthcheck()', () => {
+    it('returns response', async () => {
```
donce: The thing this test is actually testing is different - we are testing that `it sets HEALTH_CHECK_TIME metric`.

PR #213, opened 2018-06-18
feature/MYS-71 prepend daemon stderr with timestamp

shroomist:


src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -66,7 +66,7 @@ class Process {
   async setupLogging () {
     await this._prepareLogFiles()
     tailFile(this._stdoutPath, this._logCallback.bind(this, processLogLevels.INFO))
-    tailFile(this._stderrPath, this._logCallback.bind(this, processLogLevels.ERROR))
+    tailFile(this._stderrPath, prependWithTimestamp(this._logCallback.bind(this, processLogLevels.ERROR)))
```
donce: `prependWithTimestamp` name is misleading - seeing `prependWithTimestamp`, I would expect something like this:

```js
function prependWithTimestamp (data: String): String {
  return `${new Date(Date.now())}: ${data}`
}
```

That would be a single responsibility. But this class has another responsibility - decorate log callback function, which is quite hard to understand.

Can we simplify this, i.e.:
```js
async setupLogging() {
  ...
  tailFile(this._stdoutPath, this._logCallback.bind(this, processLogLevels.INFO))
  tailFile(this._stderrPath, this._logTimestampedCallback.bind(this, processLogLevels.ERROR))
  ..
}

_logTimestampedCallback (level, data) {
  const timestampedData = prependWithTimestamp(data)
  this._logCallback(timestampedData)
}
```

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -106,4 +106,10 @@ function filterLine (filter, cb) {
   }
 }
 
+function prependWithTimestamp (next) {
+  return (data) => {
+    next(`${new Date(Date.now())}: ${data}`)
+  }
+}
```
donce: Specs would allow seeing date format without launching Mysterion and trying to get STDERR from Mysterium client :D

To test this easier, we could pass in date to be able to check equality to expected string.

ignasbernotas:
MYST-601 in Jira

src/libraries/string-transform.js
```diff
@@ -0,0 +1,45 @@
...
+
+type TransformFn = (string) => string
+type Callback = (string) => void
+
+export function applyTransformation (transform: TransformFn, next: Callback): (string) => void {
+  return (data) => {
+    if (!data) return
+    next(transform(data))
+  }
+}
```
donce: Wow, this got so complicated - this is not what I had in mind at all :O

Do we really need such a complex transformation functions, when all we can have is a wrapping `._logTimestampedCallback` or something more specific, without introducing so many abstractions?

test/unit/libraries/string-transform.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+describe('applyTransformation', () => {
+  let called
+  const transformFn = (data) => {
+    called = true
+    return data + '+'
+  }
+  const cb = (data) => {
+    expect(called).to.be.true
+    expect(data).to.eql('data+')
```
donce: We are testing `applyTransformation`, so we shouldn't rely on it invoking callback - if callback doesn't get invoked, tests would still be passing, because no `expect` statements would be invoked.

Maybe `CallbackRecorder` suits this use-case?

test/unit/libraries/string-transform.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+describe('prependWith', () => {
+  it('constructs function that prepends', () => {
+    expect(prependWith('pre')('data')).to.eql('predata')
+  })
+})
+
+describe('filterByString', () => {
+  describe('constructs function that returns strings', () => {
+    it('same string if filter matches', () => {
```
donce: "it same string"?

test/unit/libraries/string-transform.spec.js
```diff
@@ -0,0 +1,54 @@
...
+import { applyTransformation, filterByString, prependWith } from '../../../src/libraries/string-transform'
+
+describe('prependWith', () => {
+  it('constructs function that prepends', () => {
+    expect(prependWith('pre')('data')).to.eql('predata')
+  })
+})
+
+describe('filterByString', () => {
+  describe('constructs function that returns strings', () => {
```
donce: This looks like a behaviour, not a describtion.

test/unit/libraries/string-transform.spec.js
```diff
@@ -0,0 +1,54 @@
...
+    return data + '+'
+  }
+  const cb = (data) => {
+    expect(called).to.be.true
+    expect(data).to.eql('data+')
+  }
+
+  it('runs transformation function and pipes its output to callback', () => {
+    applyTransformation(transformFn, cb)('data')
+  })
```
donce: There is untested special case - passing in `null` data, which is supported by implementation.

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -65,9 +66,10 @@ class Process {
 
   async setupLogging () {
     await this._prepareLogFiles()
+    const boundErrorCb = this._logCallback.bind(this, processLogLevels.ERROR)
     tailFile(this._stdoutPath, this._logCallback.bind(this, processLogLevels.INFO))
-    tailFile(this._stderrPath, this._logCallback.bind(this, processLogLevels.ERROR))
-    tailFile(SYSTEM_LOG, filterLine(INVERSE_DOMAIN_PACKAGE_NAME, this._logCallback.bind(this, processLogLevels.ERROR)))
+    tailFile(this._stderrPath, applyTransformation(prependWith(new Date(Date.now()).toString()), boundErrorCb))
```
donce: `Date.now()` will be executed only once, when `setupLogging` is invoked - all logs will be preappended with the same date.
If this is true, please add a test for this as well, because it's quite easy to miss.

test/unit/libraries/string-transform.spec.js
```diff
@@ -0,0 +1,66 @@
...
+    expect(data).to.eql('data+')
+  }
+
+  const cb2 = () => {
+    throw new Error('cb2 was called')
+  }
+
+  it('runs transformation function and pipes its output to callback', () => {
+    applyTransformation(transformOnceThenStopReturnUndefined, cb)('data')
+    applyTransformation(transformOnceThenStopReturnUndefined, cb2)('data')
```
donce: Still the same problem persists:

> We are testing `applyTransformation`, so we shouldn't rely on it invoking callback - if callback doesn't get invoked, tests would still be passing, because no `expect` statements would be invoked.
> Maybe `CallbackRecorder` suits this use-case?

src/libraries/string-transform.js
```diff
@@ -0,0 +1,47 @@
...
+export function applyTransformation (transform: TransformFn, next: Callback): (string) => void {
+  return (data) => {
+    const transformed = transform(data)
+    if (transformed !== null && transformed !== undefined) {
+      next(transformed)
+    }
+  }
+}
+
+export function prependWithFn (fn: () => string) {
```
donce: Missing return type.

src/libraries/string-transform.js
```diff
@@ -0,0 +1,47 @@
...
+  return (data: string) => {
+    return fn() + data
+  }
+}
+
+export function filterByString (filter: string) {
```
donce: Missing return type.

test/unit/libraries/string-transform.spec.js
```diff
@@ -0,0 +1,66 @@
...
+
+describe('prependWithFn', () => {
+  it('prepends each time with fn execution result', () => {
+    let calledTimes = 0
+    const prependFn = () => {
+      calledTimes += 1
+      return calledTimes.toString()
+    }
+    expect(prependWithFn(prependFn)('data')).to.eql('1data')
+    expect(prependWithFn(prependFn)('data')).to.eql('2data')
```
donce: Can we test invoking same transformation twice:
```js
const transformation = prependWithFn(prependFn)

expect(transformation('data')).to.eql('1data')
expect(transformation('data')).to.eql('2data')
````

I believe this is the exact case we encountered bug with.

File: test/unit/libraries/string-transform.spec.js [hunk 13]
```diff
@@ -0,0 +1,69 @@
...
+  })
+  it('returns nothing if filter doesn\'t match', () => {
+    expect(filterByString('a')('bold')).to.be.undefined
+  })
+})
+
+describe('applyTransformation', () => {
+  let called
+  const cbRec1 = new CallbackRecorder()
+  const cbRec2 = new CallbackRecorder()
```
donce: Why 2 CallbackRecorders for each test? We can keep 1 recorder and initialize it in `beforeEach`

test/unit/libraries/string-transform.spec.js
```diff
@@ -0,0 +1,69 @@
...
+  it('runs transformation function and pipes its output to callback', () => {
+    applyTransformation(transform, cbRec1.getCallback())('data')
+    expect(cbRec1.invoked).to.be.true
+    expect(cbRec1.argument).to.eql('data+')
+  })
+
+  it('does not call callback if transformation returns undefined or null', () => {
+    applyTransformation(transformReturnNull, cbRec2.getCallback())('data')
+    expect(cbRec2.invoked).to.be.false
+  })
```
donce: Is this expected behaviour that transformation can return `null` / `unexpected`? If not, we should throw an exception, because this can hide a bug.

test/unit/libraries/string-transform.spec.js
```diff
@@ -0,0 +1,69 @@
...
+  const cbRec1 = new CallbackRecorder()
+  const cbRec2 = new CallbackRecorder()
+  const transform = (data) => {
+    if (!called) {
+      called = true
+      return data + '+'
+    }
+  }
+  const transformReturnNull = () => {
+  }
```
donce: `called` and `transform` is only used by 1st test, and `transformReturnNull` is only used by 2nd test - moving these objects inside tests might improve readability.

test/unit/libraries/string-transform.spec.js
shroomist: `null || undefined` is an expected as a return type from transformation, in that case we don't call the `next()` callback

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -64,9 +65,11 @@ class Process {
 
   async setupLogging () {
     await this._prepareLogFiles()
+    const boundErrorCb = this._logCallback.bind(this, processLogLevels.ERROR)
+    const getTimeFn = () => new Date(Date.now()).toString()
```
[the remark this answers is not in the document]
shroomist: testing via regex now

test/unit/libraries/string-transform.spec.js
shroomist: fixed

test/unit/libraries/string-transform.spec.js
shroomist: done

src/libraries/string-transform.js
```diff
@@ -0,0 +1,58 @@
...
+  return (data: string) => {
+    if (!regex.test(data)) {
+      return
+    }
+    return data
+  }
+}
+
+function getCurrentTimeISOFormat (): string {
+  return new Date(Date.now()).toISOString()
```
donce: Isn't `Date.now()` already a Date, why do we need to pass it to `new Date`? :O

test/unit/libraries/string-transform.spec.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    applyTransformation(transformReturnNull, cbRec.getCallback())('data')
+    expect(cbRec.invoked).to.be.false
+  })
+})
+
+describe('getCurrentTimeISOFormat', () => {
+  const current = getCurrentTimeISOFormat()
+  const ISORegex = /\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)/
```
donce: Wow, that's a huge regex :O

We can avoid such regex by extracting time formatting method and testing only that:
```js
getCurrentTimeISOFormat() {
  return getTimeISOFormat(new Date(Date.now()))
}

getTimeISOFormat(date: Date) {
  return date.toISOString()
}

//spec
const date = ... // we build specific date, i.e. 2000-01-01 08:00
expect(getTimeISOFormat(date)).to.eql('2000-01-01T08:00:00.080Z')
```

This would serve as an example of format - reading regex is not that easy to understand how result looks like..

test/unit/libraries/string-transform.spec.js
```diff
@@ -0,0 +1,83 @@
...
+
+  it('runs transformation function and pipes its output to callback', () => {
+    const transform = (data) => data + '+'
+
+    applyTransformation(transform, cbRec.getCallback())('data')
+    expect(cbRec.invoked).to.be.true
+    expect(cbRec.argument).to.eql('data+')
+  })
+
+  it('does not call callback if transformation returns undefined or null', () => {
```
donce: > `null || undefined` is an expected as a return type from transformation, in that case we don't call the `next()` callback 

This hides errors from us - transformation unexpectedly and there are no warnings about that.

We should at least log unexpected behaviour. Or maybe even throw an exception?

src/libraries/string-transform.js
```diff
@@ -0,0 +1,58 @@
...
+}
+
+function prependWithFn (fn: () => string): TransformFn {
+  return (data: string) => {
+    return fn() + data
+  }
+}
+
+function filterByString (filter: string): TransformFn {
+  const regex = new RegExp(filter)
```
donce: This implementation doesn't work as you would expect from method signature:

```js
filterByString ('this is *important*')
filterByString ('+log')
```

First transformation wouldn't match `this is *important* message` string, second transformation would probably fail at all, because regex is invalid.

We can use `string.includes` for that:

```js
'Hello world, welcome to the universe.'str.includes("world");
```

src/libraries/string-transform.js
shroomist: Date.now() Returns: A Number representing the milliseconds elapsed since the UNIX epoch

test/unit/libraries/string-transform.spec.js
shroomist: I dont see the point in testing standard `Date` functionality

test/unit/libraries/string-transform.spec.js
shroomist: it is an expected behavior, if an error is thrown inside transformation then it bubbles up

src/libraries/string-transform.js
shroomist: agreed

test/unit/libraries/string-transform.spec.js
donce: Yeah, it `tail` will get that error, not us. Can we have error callback or at least log this error?

test/unit/libraries/string-transform.spec.js
donce: We already have a similar implementation:
```js
function dateTimeString (): string {
  return (new Date()).toUTCString()
}
```

If we want to have 2 different formatting functions, it would be nice to have them in the same place and to have specs on to see how each format looks like.

src/libraries/string-transform.js
```diff
@@ -0,0 +1,60 @@
...
+  return (data: string) => {
+    if (!data.includes(filter)) {
+      return
+    }
+    return data
+  }
+}
+
+function getCurrentTimeISOFormat (): string {
+  return getTimeISOFormat(new Date(Date.now()))
```
donce: `new Date(Date.now())` = `new Date()`

src/libraries/string-transform.js
donce: Thanks for clarifying :)

test/unit/libraries/strings.spec.js
```diff
@@ -0,0 +1,55 @@
...
+// @flow
+
+import { describe, it, expect } from '../../helpers/dependencies'
+import {
+  getCurrentTimeISOFormat,
+  prependWithFn,
+  toISOString
+} from '../../../src/libraries/strings'
+
+describe.only('strings', () => {
```
donce: `.only` left ❗️

src/libraries/strings.js
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// @flow
+
+type TransformFn = (string) => string
+
+function prependWithFn (fn: () => string): TransformFn {
```
donce: As discussed, namings are confusing - `filterByString` does not filter anything and `prependWithFn` does not prepend anything, they just build functions doing those actions.

`buildPrependWithFn` and `buildFilterByString`?

PR #214, opened 2018-06-18
[WIP] Feature/mys 55 backend logger winston 

shroomist:
console.log method overriding, using winston for extended control over logging

### DIFF
--- a/src/app/communication/main-communication.js
+++ b/src/app/communication/main-communication.js
@@ -25,7 +25,8 @@ import type {
   ProposalUpdateDTO,
   RequestTermsDTO,
   TermsAnsweredDTO,
-  AppErrorDTO
+  AppErrorDTO,
+  MysterionBackendLogDTO
 } from './dto'
 import type {UserSettings} from '../user-settings/user-settings'
 
@@ -43,6 +44,8 @@ interface MainCommunication {
   sendMysteriumClientUp (): void,
 
   sendMysteriumClientDown (): void,
+
+  sendMysterionMainLog (log: MysterionBackendLogDTO): void,
 
   sendProposals (proposals: ProposalUpdateDTO): void,
 
--- a/src/app/communication/messages.js
+++ b/src/app/communication/messages.js
@@ -25,6 +25,8 @@ export default {
   MYSTERIUM_CLIENT_LOG: 'mysterium-client.log',
   CURRENT_IDENTITY_CHANGED: 'current.identity.changed',
 
+  MYSTERION_MAIN_LOG: 'MYSTERION_MAIN_LOG',
+
   TERMS_REQUESTED: 'terms.requested',
   TERMS_ANSWERED: 'terms.answered',
   TERMS_ACCEPTED: 'terms.accepted',
--- a/src/app/logging/backend-logger.js
+++ b/src/app/logging/backend-logger.js
@@ -1,0 +1,75 @@
+/*
+ * Copyright (C) 2018 The "MysteriumNetwork/mysterion" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// @flow
+
+import winston from 'winston'
+import LogCache from './log-cache'
+import { MainLogCachingTransport, MainLogCommunicationTransport } from './backend-logging-transports'
+import type { MainCommunication } from '../communication/main-communication'
+
+interface WinstonLogger {
+  info: Function,
+  warn: Function,
+  error: Function,
+  debug: Function,
+  add: Function
+}
+
+export default class BackendLogger {
+  _logger: WinstonLogger
+  backendLogCache: LogCache
+  mysteriumProcessLogCache: LogCache
+
+  constructor (backendLogCache: LogCache, mysteriumProcessLogCache: LogCache) {
+    this.backendLogCache = backendLogCache
+    this.mysteriumProcessLogCache = mysteriumProcessLogCache
+
+    this._logger = winston.createLogger({
+      transports: [
+        new winston.transports.Console(),
+        new MainLogCachingTransport(this.backendLogCache)
+      ]
+    })
+  }
+
+  overrideConsoleLogs () {
+    // $FlowFixMe
+    console.log = (...args) => this._logger.info(args.join(' '))
+    // $FlowFixMe
+    console.info = (...args) => this._logger.info(args.join(' '))
+    // $FlowFixMe
+    console.warn = (...args) => this._logger.warn(args.join(' '))
+    // $FlowFixMe
+    console.error = (...args) => this._logger.error(args.join(' '))
+    // $FlowFixMe
+    console.debug = (...args) => this._logger.debug(args.join(' '))
+  }
+
+  sendCachedViaCommunication (com: MainCommunication) {
+    for (const infoEntry of this.backendLogCache.get().info) {
+      com.sendMysterionMainLog(infoEntry)
+    }
+    for (const errorEntry of this.backendLogCache.get().error) {
+      com.sendMysterionMainLog(errorEntry)
+    }
+  }
+
+  addMainCommunicationTransport (com: MainCommunication) {
+    this._logger.add(new MainLogCommunicationTransport(com))
+  }
+}
--- a/src/app/logging/backend-logging-transports.js
+++ b/src/app/logging/backend-logging-transports.js
@@ -1,0 +1,63 @@
+/*
+ * Copyright (C) 2018 The "MysteriumNetwork/mysterion" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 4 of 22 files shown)


src/app/communication/main-communication.js
```diff
@@ -44,6 +45,8 @@ interface MainCommunication {
 
   sendMysteriumClientDown (): void,
 
+  sendMysterionMainLog (log: MysterionBackendLogDTO): void,
```
donce: Different names for the same thing - *Mysterion main*, *Mysterion Backend* - we should use single term for the same thing.

src/app/logging/backend-logger.js
```diff
@@ -0,0 +1,75 @@
...
+import { MainLogCachingTransport, MainLogCommunicationTransport } from './backend-logging-transports'
+import type { MainCommunication } from '../communication/main-communication'
+
+interface WinstonLogger {
+  info: Function,
+  warn: Function,
+  error: Function,
+  debug: Function,
+  add: Function
+}
```
donce: `flow-typed` exposes some kind of interfaces for `winston`, can we re-use them? https://github.com/flowtype/flow-typed/blob/master/definitions/npm/winston_v3.x.x/flow_v0.34.x-/winston_v3.x.x.js

src/app/logging/backend-logger.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+    for (const errorEntry of this.backendLogCache.get().error) {
+      com.sendMysterionMainLog(errorEntry)
+    }
+  }
+
+  addMainCommunicationTransport (com: MainCommunication) {
+    this._logger.add(new MainLogCommunicationTransport(com))
+  }
+}
```
donce: `BackendLogger` logger has 2 main responsibilities:
- logging to multiple sources.
- overriding `console` methods.

We should separate some responsibilities somehow, i.e.:

```js
class BackendLogger {
  ...

  log () {
    this._logger.info(..)
  }

  info () {
    this._logger.info(..)
  }

  startSync (communication: MainCommunication) {
    this._ sendCachedViaCommunication(communication)
    this._ addMainCommunicationTransport(communication)
  }

  _sendCachedViaCommunication(...) {
    ...
  }

  _addMainCommunicationTransport(...) {
    ...
  }
...  
}



//separate function
function overrideConsole (logger: BackendLogger) {
  console.log = logger.log
  console.info = logger.info
  ...
}
```

src/app/logging/backend-logger.js
```diff
@@ -0,0 +1,75 @@
...
+      com.sendMysterionMainLog(infoEntry)
+    }
+    for (const errorEntry of this.backendLogCache.get().error) {
+      com.sendMysterionMainLog(errorEntry)
+    }
+  }
+
+  addMainCommunicationTransport (com: MainCommunication) {
+    this._logger.add(new MainLogCommunicationTransport(com))
+  }
```
donce: Hmm, maybe we can hide communication inside `MainLogCachingTransport`? At first, it can cache everything, but once communication is available, we can invoke `MainLogCachingTransport.startSync`, which would flush all cached logs and will push new logs to communication.

src/app/logging/backend-log-setup.js
```diff
@@ -0,0 +1,74 @@
...
+  // $FlowFixMe
+  console.log = (...args) => logger.info(args.join(' '))
+  // $FlowFixMe
+  console.info = (...args) => logger.info(args.join(' '))
+  // $FlowFixMe
+  console.warn = (...args) => logger.warn(args.join(' '))
+  // $FlowFixMe
+  console.error = (...args) => logger.error(args.join(' '))
+  // $FlowFixMe
+  console.debug = (...args) => logger.debug(args.join(' '))
```
donce: If we're using `console` for logging, we should replace `logger` usage with `console`.

src/app/logging/backend-logging-transports.js
```diff
@@ -0,0 +1,63 @@
...
+  message: string
+}
+
+export class BackendLogCommunicationTransport extends Transport {
+  constructor (com: MainCommunication) {
+    super()
+    this.communication = com
+  }
+
+  log (info: LogEntry, callback: Function) {
```
donce: Can we specify `callback` type `() => any`?

src/app/logging/backend-logging-transports.js
```diff
@@ -0,0 +1,63 @@
...
+import type { MainCommunication } from '../communication/main-communication'
+import LogCache from './log-cache'
+
+type LogEntry = {
+  level: string,
+  message: string
+}
+
+export class BackendLogCommunicationTransport extends Transport {
+  constructor (com: MainCommunication) {
```
donce: Since communication is called `MainCommunication` and this is just transport for that, can't we just call it `MainCommunicationTransport` to have fewer different terms?

src/app/logging/backend-logging-transports.js
```diff
@@ -0,0 +1,63 @@
...
+  log (info: LogEntry, callback: Function) {
+    this.communication.sendMysterionBackendLog(info)
+    callback()
+  }
+}
+
+export class BackendLogCachingTransport extends Transport {
+  constructor (logCache: LogCache) {
+    super()
+    this.logCache = logCache
```
donce: Property for `logCache` is missing - Flow is probably not complaining, because it doesn't understand super class `Transport`.
Also, it should be prefixed with `_`.

src/app/logging/backend-logging-transports.js
```diff
@@ -0,0 +1,63 @@
...
+
+type LogEntry = {
+  level: string,
+  message: string
+}
+
+export class BackendLogCommunicationTransport extends Transport {
+  constructor (com: MainCommunication) {
+    super()
+    this.communication = com
```
donce: Property for `communication` is missing - Flow is probably not complaining, because it doesn't understand super class `Transport`.
Also, it should be prefixed with `_`.

src/app/logging/log-cache.js
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017 The "MysteriumNetwork/mysterion" Authors.
+ * Copyright (C) 2018 The "MysteriumNetwork/mysterion" Authors.
```
donce: Hmm, does WebStorm update all dates automatically? We shouldn't do that - this date should be creation date, not last modified date.

src/main/dependencies/modules/bug-reporting.js
```diff
@@ -19,19 +19,30 @@
...
 import type {Container} from '../../../app/di'
+import BackendLogSetup, { overrideConsoleLogs } from '../../../app/logging/backend-log-setup'
 
 function bootstrap (container: Container) {
+  container.factory(
+    'backendLogger',
+    ['backendLogCache', 'mysteriumProcessLogCache'],
+    (backendLogCache, mysteriumProcessLogCache) => {
+      const backendLogger = new BackendLogSetup(backendLogCache, mysteriumProcessLogCache)
+      overrideConsoleLogs(backendLogger.winstonLogger)
```
donce: This huge side-effect is hidden inside DI - it should be done explicitly, i.e. in `Mysterion` class.

src/app/communication/messages.js
```diff
@@ -25,6 +25,8 @@ export default {
   MYSTERIUM_CLIENT_LOG: 'mysterium-client.log',
   CURRENT_IDENTITY_CHANGED: 'current.identity.changed',
 
+  MYSTERION_BACKEND_LOG: 'MYSTERION_MAIN_LOG',
```
donce: Please unify these.

src/app/logging/backend-log-setup.js
```diff
@@ -0,0 +1,88 @@
...
+
+  debug (string): void,
+
+  add (Transport): void
+}
+
+export default class BackendLogSetup {
+  winstonLogger: Logger
+  backendLogCache: LogCache
+  mysteriumProcessLogCache: LogCache
```
donce: These doesn't seem to be used outside of class, lets make these private to make interface easier to understand.

src/app/logging/backend-logging-transports.js
```diff
@@ -0,0 +1,67 @@
...
+
+export class BackendLogCommunicationTransport extends Transport {
+  _communication: MainCommunication
+  constructor (com: MainCommunication) {
+    super()
+    this._communication = com
+  }
+
+  log (info: LogEntry, callback: Function) {
+    this._communication.sendMysterionBackendLog({ level: mapToCacheLevel(info.level), message: info.message })
```
donce: Hmm, so if we'll use `warn` level, it will be `warn` in `Backend` but `info` in `Renderer`?
Also, very confusing name - not sure what you're mapping, because `this._communication.sendMysterionBackendLog` doesn't say anything about cache - why do we have to invoke mapping to cache level?

src/app/logging/backend-logging-transports.js
```diff
@@ -0,0 +1,67 @@
...
+
+function mapToCacheLevel (level: string): 'info' | 'error' {
+  switch (level) {
+    case 'error':
+      return 'error'
+    case 'info':
+    case 'warn':
+    case 'verbose':
+    case 'debug':
+    case 'silly':
```
donce: we doesn't seem to expose `silly` log level anywhere, why do we need to know about that?
This is what we expose:
```js
+  info (string): void,
+
+  warn (string): void,
+
+  error (string): void,
+
+  debug (string): void,
```

src/app/logging/backend-logging-transports.js
```diff
@@ -47,20 +47,18 @@ export class BackendLogCachingTransport extends Transport {
...
 
   log (info: LogEntry, callback: () => any) {
-    this._logCache.pushToLevel(mapToCacheLevel(info.level), info.message)
+    this._logCache.pushToLevel(mapToLogLevel(info.level), info.message)
     callback()
   }
 }
 
-function mapToCacheLevel (level: string): 'info' | 'error' {
+function mapToLogLevel (level: string): 'info' | 'error' {
```
donce: Still very confusing - it accepts `level` argument and maps it to `level`? Could you improve naming or add a TODO to explain what domains are this map between?

PR #217, opened 2018-06-21
Hotfix/log format improvement

shroomist:
removing flow-typed libs, as they were not precise.
specifically `$winstonFormatSubModule` was missing `printf` function, that is being used here


shroomist:
it would be good to have timestamps @tadovas, perhaps you'd like to approve this? ^_^

PR #222, opened 2018-06-28
MYS-68 Add callbacks removal, remove onUserSettings callback

donce:


src/app/communication/rendererMessageBus.js
```diff
@@ -17,16 +17,30 @@
...
-  on (channel: string, callback: Function): void {
-    ipcRenderer.on(channel, (event, data) => {
+
+  on (channel: string, callback: MessageBusCallback): void {
+    const listener = (event, data) => {
       callback(data)
-    })
+    }
+    ipcRenderer.on(channel, listener)
+    this._listeners[callback] = listener
```
ignasbernotas: do we really need a listener storage here? seems like it's only use is checking whether it can be removed:)

test/unit/renderer/components/DisconnectNotificationSetting.spec.js
```diff
@@ -0,0 +1,53 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {createLocalVue, mount} from '@vue/test-utils'
+import RendererCommunication from '../../../../src/app/communication/renderer-communication'
+import DIContainer from '../../../../src/app/di/vue-container'
+import FakeMessageBus from '../../../helpers/fakeMessageBus'
+import DisconnectNotificationSetting from '@/components/DisconnectNotificationSetting'
+
+function mountWith (rendererCommunication, store) {
```
ignasbernotas: Maybe add a todo to extract this? Not sure if it's worth it, but we've got quite a few tests that have the same function.

donce: Added a TODO.

test/helpers/utils.js
```diff
@@ -55,13 +55,9 @@ async function captureAsyncError (func: () => Promise<any>) {
...
-  invoked: boolean
-  _argument: any
-
-  constructor () {
-    this.invoked = false
-    this._argument = null
-  }
+  invoked: boolean = false
+  _argument: any = null
+  _bindedCallback: (any) => void
```
shroomist: engrish please `boundCallback` 🇬🇧

test/helpers/subscribable-message-bus.js
```diff
@@ -0,0 +1,55 @@
...
+  on (channel: string, callback: MessageBusCallback): void {
+    this._callbacks[channel] = callback
+    this._callbacksCount++
+  }
+
+  removeCallback (channel: string, callback: MessageBusCallback): void {
+    if (this._callbacks[channel] !== callback) {
+      throw new Error('Callback being removed was not found')
+    }
+    this._callbacks[channel] = undefined
```
shroomist: I'd prefer you doing `delete this._callbacks[channel]`

src/app/communication/rendererMessageBus.js
```diff
@@ -17,16 +17,48 @@
...
+    this._registerListener(channel, listener)
+  }
+
+  removeCallback (channel: string, callback: MessageBusCallback): void {
+    const listener = this._listeners[callback]
+    if (!listener) {
+      throw new Error(`Removing callback for '${channel}' message in renderer failed - callback was not found`)
+    }
+    this._removeListener(channel, listener)
+    this._listeners[callback] = undefined
```
shroomist: I'd prefer you doing `delete this._callbacks[channel]`
you'd able to do a more precise mapping `_listeners: { [MessageBusCallback]:` ~~?~~ `Listener`

src/app/communication/rendererMessageBus.js
```diff
@@ -17,16 +17,48 @@
...
+
   send (channel: string, data?: mixed): void {
     ipcRenderer.send(channel, data)
   }
-  on (channel: string, callback: Function): void {
-    ipcRenderer.on(channel, (event, data) => {
+
+  on (channel: string, callback: MessageBusCallback): void {
+    const listener = this._buildListener(callback)
+    this._listeners[callback] = listener
```
shroomist: can we test the dirty case, when we try to subscribe same callback multiple times
wether or not it would be notified multiple times when event occurs?

test/helpers/subscribable-message-bus.js
donce: Yes, `= undefined` looks like a hack, but indexer does not support `delete`: https://flow.org/en/docs/types/objects/#toc-objects-as-maps

src/app/communication/rendererMessageBus.js
donce: Yes, `= undefined` looks like a hack, but indexer does not support `delete`: https://flow.org/en/docs/types/objects/#toc-objects-as-maps

test/helpers/utils.js
donce: Fixed 👨🏼‍✈️

src/app/communication/rendererMessageBus.js
donce: Unit-testing this case of `RendererMessageBus` would invoke `electron` methods (`ipcRenderer.on`), which is  not available in unit-tests. We might extract out one more abstraction to test this out, but I'm afraid it will make communication even harder to understand.

What are your thoughts @shroomist ?

test/helpers/subscribable-message-bus.js
shroomist: don't see what you mean, just tested `delete this._callbacks[channel]` and all checks passed

donce: Cool, it works - I think I've only tried using method, not the operator. Fixed, thanks :)

src/app/communication/rendererMessageBus.js
donce: Fixed.

src/app/communication/rendererMessageBus.js
shroomist: sure, it would be awesome if we could pass any `PubSub` class as a transport here.

PR #223, opened 2018-06-28
Add integration tests to CI

donce:
`yarn flow` give instant results if running it multiple times - it's very convenient to run `yarn flow` as first step when developing because errors are shown instantly.
Also, integration tests were skipped in CI for some reasons - it doesn't have to be skipped anymore.


donce:
@shroomist there is no `CONTRIBUTING.md` yet, it's in another PR.

PR #224, opened 2018-06-28
Add eslint-plugin-filenames to see existing naming inconsistencies

donce:
Rule is disabled with a TODO. By uncommenting this rule locally, we get what files need to be renamed before we can enable this rule in CI.


donce:
@shroomist 
>so for now, does it warn on wrong filenames?

for now, it doesn't warn on wrong filenames - it warns only if you uncomment single line in "rules" section.

> what changes after TODO is fixed?

After all existing filenames are unified, we can re-enable rule which will keep future names consistent.

> I suggest we fix TODO: fix existing filename warnings
in this PR

It's risky to do all renamings in a single PR - it might cause many annoying git conflicts. That's why I want to do that gradually - and this rule shows what's remaining.

But it seems alright to merge this in only after all renamings are done.

PR #226, opened 2018-06-29
Hotfix/timestamp transports

shroomist:
timestamp log messages before putting into LogCache or sending via Communication


src/app/logging/backend-logging-transports.js
```diff
@@ -35,8 +36,10 @@ export class SyncCommunicationTransport extends Transport {
     this._communication = communication
   }
 
-  log (info: LogEntry, callback: () => any) {
-    const logDto = { level: mapToLogLevel(info.level), data: info.message }
+  log (logEntry: LogEntry, callback: () => any) {
+    const timestamp = logEntry.timestamp || ''
+    const message = timestamp + logEntry.message
```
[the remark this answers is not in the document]
shroomist: @donce opinions on this?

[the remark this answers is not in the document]
donce: How does timestamp looks like? Maybe we can shorten timestamp string to save one byte instead of skipping separator?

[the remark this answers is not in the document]
shroomist: `2018-06-29T12:42:48.969Z` <- Z denotes the zero timezone.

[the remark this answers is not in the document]
donce: Whichever seems fine for me - with space or without space :D

src/app/logging/backend-logging-transports.js
```diff
@@ -24,7 +24,8 @@ import type { LogLevel } from './index'
 
 type LogEntry = {
   level: string,
-  message: string
+  message: string,
+  timestamp?: string
 }
 
 export class SyncCommunicationTransport extends Transport {
```
donce: Just realised that this transport is only used in renderer, so it's confusing keeping this class in `backend-logging-transports.js` file. This is not related to this PR and it's probably my doing, but could you add a TODO for this not to forget about that?

src/app/logging/backend-logging-transports.js
```diff
@@ -49,8 +52,10 @@ export class BackendLogCachingTransport extends Transport {
     this._logCache = logCache
   }
 
-  log (info: LogEntry, callback: () => any) {
-    this._logCache.pushToLevel(mapToLogLevel(info.level), info.message)
+  log (logEntry: LogEntry, callback: () => any) {
+    const timestamp = logEntry.timestamp || ''
+    const message = timestamp + logEntry.message
```
donce: This formatting is duplicated for both transports, we should extract it.
Also, not quite sure why the current formatting isn't applied in these transports? It's in `log-bootstrapping.js`:
```js
const winstonFormat = winston.format.combine(
  winston.format.timestamp(),
  winston.format.printf(log => `${log.timestamp} ${log.level}: ${log.message}`)
)
```

shroomist: the new formatting is added to `logEntry` object as a `Symbol` and should be accessed accordingly.
screen shot 2018-07-02 at 12 51 15

shroomist: I prefer constructing message string myself here

src/app/logging/backend-logging-transports.js
shroomist: @donce refactored this

src/app/logging/winston-log-entry.js
```diff
@@ -0,0 +1,41 @@
...
+
+import type { LogLevel } from './mysterion-log-levels'
+
+type WinstonLogEntry = {
+  level: string,
+  message: string,
+  timestamp?: string
+}
+
+function mapWinstonLogLevelToMysterionLevel (level: string): LogLevel {
```
donce: This clarifies mapping a lot! 💯

src/app/logging/frontend-log-transport.js
```diff
@@ -18,51 +18,25 @@
...
   constructor (communication: SyncRendererCommunication) {
     super()
     this._communication = communication
   }
 
-  log (info: LogEntry, callback: () => any) {
-    const logDto = { level: mapToLogLevel(info.level), data: info.message }
+  log (logEntry: WinstonLogEntry, callback: () => any) {
+    const timestamp = logEntry.timestamp || ''
+    const message = timestamp + logEntry.message
```
donce: This logic of formatting WinstonLogEntry to string is duplicated in `BackendLogCachingTransport.log`, we should extract it and re-use.

PR #231, opened 2018-07-02
Added flow support to mysterium_client lib and fixed MYS-73

ignasbernotas:


src/libraries/mysterium-client/launch-daemon/installer.js
```diff
@@ -32,13 +36,11 @@ function processInstalled () {
...
 }
 
-class Installer {
-  /**
-   * @constructor
-   * @param {ClientConfig} config
-   */
-  constructor (config) {
-    this.config = config
+class Installer implements InstallerInterface {
```
donce: `InstallerInterface` for interface name looks redundant, while `Installer` for class name doesn't look specific enough.
`class LaunchDaemonInstaller implements Installer` ?

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -47,8 +57,8 @@ class Process {
     }
   }
 
-  start () {
-    return axios.get('http://127.0.0.1:' + this._daemonPort)
+  async start (): Promise<void> {
+    axios.get('http://127.0.0.1:' + this._daemonPort)
```
donce: This `async` method is not actually waiting for `axios.get` result.

If you want it to wait for request finish:
- We should keep `return` or use `try { await .. } catch (err) {}` syntax.

If you don't want to wait for request finish, you can just remove `async` from method signature.

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -29,15 +33,21 @@ const stderrFileName = 'stderr.log'
 /**
  * Spawns and stops 'mysterium_client' daemon on OSX
  */
-class Process {
+class Process implements ProcessInterface {
+  _tequilapi: TequilapiClient
+  _daemonPort: number
+  _stdoutPath: string
+  _stderrPath: string
+  _subscribers: Object
```
donce: `Object` is still dangerous, we should be more specific - we can use indexer (map) here.

src/libraries/mysterium-client/index.js
```diff
@@ -15,30 +15,26 @@
...
-  default:
-    Installer = StandaloneInstaller
-    // TODO: fix this, because StandaloneProcess constructor differs from LaunchDaemonProcess constructor, but they are
-    // constructed with same parameters
-    Process = StandaloneProcess
-    break
+interface InstallerInterface {
+  needsInstallation (): boolean,
+
+  install (): Promise<any>
```
donce: `Promise<void>`?

src/libraries/mysterium-client/standalone/installer.js
```diff
@@ -15,12 +15,18 @@
...
+import type { InstallerInterface } from '../index'
+
+class Installer implements InstallerInterface {
+  needsInstallation (): boolean {
     return false
   }
 
-  install () {}
+  install (): Promise<void> {
+    return Promise.resolve()
```
donce:
```js
async install (): Promise<void> {
}
```

src/libraries/mysterium-client/standalone/process.js
```diff
@@ -15,68 +15,80 @@
...
 
+type LogCallback = (data: any) => any
+
 /**
  * 'mysterium_client' process handler
  */
-class Process {
+class Process implements ProcessInterface {
+  _config: ClientConfig
+  _child: ChildProcess
```
donce: type should be `?ChildProcess`, because it can be null.

src/libraries/mysterium-client/index.js
```diff
@@ -15,30 +15,26 @@
...
+}
+
+interface ProcessInterface {
+  start (): Promise<any>,
+
+  stop (): Promise<any>,
+
+  onLog (level: string, callback: Function): void,
```
donce: Callback function has already defined type: `LogCallback` in `standalone/process.js`

src/libraries/mysterium-client/launch-daemon/process.js
```diff
@@ -57,24 +67,29 @@ class Process {
...
     await this._prepareLogFiles()
+
     tailFile(this._stdoutPath, this._logCallback.bind(this, processLogLevels.INFO))
     tailFile(this._stderrPath, this._logCallback.bind(this, processLogLevels.ERROR))
     tailFile(SYSTEM_LOG, filterLine(INVERSE_DOMAIN_PACKAGE_NAME, this._logCallback.bind(this, processLogLevels.ERROR)))
   }
 
-  onLog (level, cb) {
-    if (!this._subscribers[level]) throw new Error(`Unknown process logging level: ${level}`)
+  onLog (level: string, cb: Function): void {
```
donce: Callback function has already defined type: `LogCallback` in `standalone/process.js`

src/libraries/mysterium-client/index.js
```diff
@@ -15,30 +15,26 @@
...
-    Process = LaunchDaemonProcess
-    break
-
-  default:
-    Installer = StandaloneInstaller
-    // TODO: fix this, because StandaloneProcess constructor differs from LaunchDaemonProcess constructor, but they are
-    // constructed with same parameters
-    Process = StandaloneProcess
-    break
+interface InstallerInterface {
```
shroomist: perhaps there's no need to prepend interface name with `Interface`?

src/libraries/mysterium-client/launch-daemon/process.js
shroomist:
```
{
  info: [SubstcribeFunctions],
  error: [SubscribeFunctions]
 }

File: src/libraries/mysterium-client/standalone/process.js (hunk 6 above)
Comment (shroomist) [reply src/libraries/mysterium-client/standalone/process.js]: @donce you keep mixing `null` and `undefined`.
it certainly can be `undefined`, there's no code that makes it `null`

Comment (donce) [reply src/libraries/mysterium-client/standalone/process.js]: Yeah - it can be `undefined` after construction, so it should be `?ChildProcess`

File: src/libraries/mysterium-client/index.js (hunk 9 above)
Comment (donce) [reply src/libraries/mysterium-client/index.js]: Agree - as in my previous comment, it could be just `Installer`, and implementation classes could be longer, i.e. `StandaloneInstaller implements Installer`.

File: src/libraries/mysterium-client/launch-daemon/installer.js (hunk 1 above)
Comment (ignasbernotas) [reply src/libraries/mysterium-client/launch-daemon/installer.js]: Fixed

File: src/libraries/mysterium-client/launch-daemon/process.js (hunk 2 above)
Comment (ignasbernotas) [reply src/libraries/mysterium-client/launch-daemon/process.js]: Fixed

File: src/libraries/mysterium-client/launch-daemon/process.js (hunk 3 above)
Comment (ignasbernotas) [reply src/libraries/mysterium-client/launch-daemon/process.js]: Fixed

File: src/libraries/mysterium-client/index.js (hunk 4 above)
Comment (ignasbernotas) [reply src/libraries/mysterium-client/index.js]: Fixed

File: src/libraries/mysterium-client/standalone/installer.js (hunk 5 above)
Comment (ignasbernotas) [reply src/libraries/mysterium-client/standalone/installer.js]: Fixed

File: src/libraries/mysterium-client/standalone/process.js (hunk 6 above)
Comment (ignasbernotas) [reply src/libraries/mysterium-client/standalone/process.js]: Fixed

File: src/libraries/mysterium-client/index.js (hunk 7 above)
Comment (ignasbernotas) [reply src/libraries/mysterium-client/index.js]: Fixed

File: src/libraries/mysterium-client/launch-daemon/process.js (hunk 8 above)
Comment (ignasbernotas) [reply src/libraries/mysterium-client/launch-daemon/process.js]: Fixed

File: src/libraries/mysterium-client/index.js (hunk 9 above)
Comment (ignasbernotas) [reply src/libraries/mysterium-client/index.js]: Fixed

File: src/libraries/mysterium-client/launch-daemon/process.js [hunk 10]
```diff
@@ -57,24 +71,29 @@ class Process {
...
+  onLog (level: string, cb: LogCallback): void {
+    if (!this._subscribers[level]) {
+      throw new Error(`Unknown process logging level: ${level}`)
+    }
+
     this._subscribers[level].push(cb)
   }
 
-  _logCallback (level, data) {
+  _logCallback (level: string, data: mixed) {
```
donce: level can be more strict - we can re-use `processLogLevels.INFO | processLogLevels.ERROR` from `Subscribers`, i.e.

```js
type LogLevels = processLogLevels.INFO | processLogLevels.ERROR

type Subscribers = {
  [LogLevels]: Array<LogCallback>,
}

...
_logCallback (level: LogCallback, data: mixed) {
```

This can be applied in multiple other places as well.

src/libraries/mysterium-client/index.js
```diff
@@ -15,30 +15,28 @@
...
+interface Installer {
+  needsInstallation (): boolean,
+
+  install (): Promise<void>
+}
+
+interface Process {
+  start (): void,
+
+  stop (): Promise<any>,
```
donce: It's confusing that `start` is sync, but `stop` is async - we should keep them consistent, both sync or both async.

src/main/dependencies/modules/mysterium-client.js
```diff
@@ -16,15 +16,30 @@
...
+import type { ClientConfig } from '../../../libraries/mysterium-client/config'
 import type { TequilapiClient } from '../../../libraries/mysterium-tequilapi/client'
 import { LAUNCH_DAEMON_PORT } from '../../../libraries/mysterium-client/launch-daemon/config'
+import os from 'os'
 
 function bootstrap (container: Container) {
+  container.service(
+    'mysteriumClient.platform',
+    [],
+    () => {
```
donce: can be just `() => os.platform()`

src/main/dependencies/modules/mysterium-client.js
```diff
@@ -42,14 +55,26 @@ function bootstrap (container: Container) {
   )
   container.service(
     'mysteriumClientInstaller',
-    ['mysteriumClient.config'],
-    (config) => new Installer(config)
+    ['mysteriumClient.config', 'mysteriumClient.platform'],
+    (config: ClientConfig, platform: string) => {
+      switch (platform) {
```
shroomist: earlier, I think it was cool that we only needed one `switch` controlling logic for different platforms.
now I'm concerned that this is starting to spread.

In fact I did like that `library` was taking care of choosing the right `installer` and `process` handlers for us, and implementation didn't have to worry about platform.
thoughts? @donce @Waldz

ignasbernotas: The problem is that even though the library chose which handlers to use it didn't solve the issue. Even with the right handlers, you would still need a separate switch for the platform since the constructors are different so the previous solution didn't work and would still need to be rewritten. My PR fixes this issue in app level rather than library level and works well now. Once we decide to move the lib into an npm package we can rethink how it could be handled, but for now I think my approach is more beneficial to us.

PR #233, opened 2018-07-03
Fix ProposalFetcher.stop error when rejecting terms

donce:
When rejecting terms, error is printed into node console, saying that `stop` method is not available on `undefined`. This is because `ProposalFetcher.start()` was not invoked.


src/app/data-fetchers/proposal-fetcher.js
```diff
@@ -39,14 +39,14 @@ class ProposalFetcher {
    * Starts periodic proposal fetching.
    */
   start (): void {
-    this._loop = new FunctionLooper(async () => {
+    const loop = new FunctionLooper(async () => {
```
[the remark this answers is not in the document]
donce: I did this initially, but than Flow throws an exception than it can be `null`- not totally sure, but I'm guessing that  Flow does not know everything about class properties, so he just always assumes it can be null.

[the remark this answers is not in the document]
ignasbernotas: Yeah, same happened to me with ChildProcess prop:)

src/app/data-fetchers/proposal-fetcher.js
```diff
@@ -39,14 +39,14 @@ class ProposalFetcher {
    * Starts periodic proposal fetching.
    */
   start (): void {
```
mipo47: Add validation for double start, example:
if (this._loop) {
  return
}

donce: Good catch, we can fix this as well, but in another PR, since this PR is just about `stop` and it's already being reviewed 6 days :D

donce: Added tech dept item for that not to forget about it: https://mysteriumnetwork.atlassian.net/browse/MYS-124

src/libraries/functionLooper.js
```diff
@@ -31,9 +31,9 @@ class FunctionLooper {
   _threshold: number
   _running: boolean = false
   _stopping: boolean = false
-  _currentExecutor: ThresholdExecutor
-  _currentPromise: Promise<void>
   _errorSubscriber: Subscriber<Error> = new Subscriber()
+  _currentExecutor: ?ThresholdExecutor
```
shroomist: how didn't we catch this optional earlier..?

donce: Living and learning from mistakes :D A bit sad that Flow doesn't enforce this to be `maybe` type, even though it's not set in constructor..

PR #234, opened 2018-07-03
Refactor/mys 100 unify logging

shroomist:
removed console.log overrides from backend. refactored logging


src/renderer/dependencies/modules/bug-reporting.js
```diff
@@ -27,17 +27,17 @@ import type { EnvironmentCollector } from '../../../app/bug-reporting/environmen
...
-import FrontendLogBootstrapper from '../../../app/logging/frontend-log-bootstrapper'
+import { createWinstonSyncComLogger } from '../../../app/logging/winston'
 import { BugReporterMetrics } from '../../../app/bug-reporting/bug-reporter-metrics'
 
 function bootstrap (container: Container) {
   container.constant('bugReporter.sentryURL', 'https://f1e63dd563c34c35a56e98aa02518d40@sentry.io/300978')
 
   container.factory(
-    'frontendLogBootstrapper',
+    'stringLogger',
```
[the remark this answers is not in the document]
shroomist: I'd actually prefer this to be named `cachingLogger` for both front and backend.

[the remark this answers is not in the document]
donce: Yeah, `stringLogger` confuses a lot :D
Just `logger` sounds good for me - `cachingLogger` seems ok too, but it might cause some confusion, because in frontend, logger doesn't know anything about caching, it just passes it to main and that's all (caching is implementation detail of backend).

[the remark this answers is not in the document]
donce: Just not sure about the `StringLogger` interface - renaming that to `Logger` would cause name clashing with `Logger` class. Maybe it's fine to call that interface `WinstonLogger`? It exposes implementation, but it is what it is - without winston, we wouldn't need that interface. What do you think @shroomist ?

src/app/logging/backend-log-bootstrapper.js
```diff
@@ -1,58 +0,0 @@
...
-  // $FlowFixMe
-  console.log = (...args) => logger.info(args.join(' '))
-  // $FlowFixMe
-  console.info = (...args) => logger.info(args.join(' '))
-  // $FlowFixMe
-  console.warn = (...args) => logger.warn(args.join(' '))
-  // $FlowFixMe
-  console.error = (...args) => logger.error(args.join(' '))
-  // $FlowFixMe
-  console.debug = (...args) => logger.debug(args.join(' '))
```
donce: There are still some occurrences of `console` left, we should probably remove those and add `no-console` linter rule to keep logging consistent.

src/libraries/mysterium-client/launch-daemon/launch-daemon-process.js
```diff
@@ -63,19 +63,11 @@ class LaunchDaemonProcess implements Process {
   }
 
   async start (): Promise<void> {
-    try {
-      await axios.get('http://127.0.0.1:' + this._daemonPort)
-
-      console.info('Touched the daemon, now it should be up')
-    } catch (e) {
-      console.info('Touched the daemon with error, anyway it should be up')
```
donce: These logs seem useful - maybe we can keep them where `.start()` and `.stop()` is invoked?

src/app/logger.js
```diff
@@ -17,34 +17,22 @@
...
 // @flow
 
-import type { StringLogger } from './logging/log-boostrapping'
+import type { StringLogger } from './logging/string-logger'
 
-/**
- * Used as logging entrypoint in frontend
- */
-// TODO: unify how logging is done in backend and frontend (MYS-100)
 class Logger {
```
mipo47: Logger class only converts array to string, maybe we can remove this class? (ex: move join(' ') to StringLogger

src/app/logging/backend-log-bootstrapper.js
mipo47: I would recommend to leave console overrides, but show warning: console shouldn't be used anymore (maybe send warning to bugreporter)

src/app/logging/frontend-log-bootstrapper.js
```diff
@@ -1,45 +0,0 @@
...
-
-// @flow
-
-import winston from 'winston'
-import type { SyncRendererCommunication } from '../communication/sync/sync-communication'
-import SyncCommunicationTransport from './frontend-log-transport'
-import logger from '../logger'
-import { winstonFormat } from './log-boostrapping'
-
-class FrontendLogBootstrapper {
```
mipo47: Recommend to use DI for class creation (put code to folder dependencies)

src/app/logging/winston-log-entry.js
```diff
@@ -1,41 +0,0 @@
...
-import type { LogLevel } from './mysterion-log-levels'
-
-type WinstonLogEntry = {
-  level: string,
-  message: string,
-  timestamp?: string
-}
-
-function mapWinstonLogLevelToMysterionLevel (level: string): LogLevel {
-  switch (level) {
```
mipo47: return level == 'error' ? 'error' : 'info'

src/app/logging/winston.js
```diff
@@ -0,0 +1,75 @@
...
+  return winstonLogger
+}
+
+function createWinstonSyncComLogger (communication: SyncRendererCommunication) {
+  const winstonLogger = createWinstonLogger()
+  winstonLogger.add(new WinstonTransportSyncCom(communication))
+  return winstonLogger
+}
+
+function mapWinstonLogLevelToMysterionLevel (level: string): LogLevel {
```
mipo47: Reuse same function from src/app/logging/winston-log-entry.js

src/app/mysterion.js
```diff
@@ -60,7 +61,7 @@ type MysterionParams = {
   bugReporter: BugReporter,
   environmentCollector: EnvironmentCollector,
   bugReporterMetrics: BugReporterMetrics,
-  backendLogBootstrapper: BackendLogBootstrapper,
+  stringLogger: StringLogger,
```
mipo47: stringLogger used only by logger (logger.setLogger(this.stringLogger)), 
maybe it's possible to pass logger here instead of stringLogger

src/libraries/mysterium-client/launch-daemon/launch-daemon-process.js
```diff
@@ -65,17 +65,13 @@ class LaunchDaemonProcess implements Process {
   async start (): Promise<void> {
     try {
       await axios.get('http://127.0.0.1:' + this._daemonPort)
-
-      console.info('Touched the daemon, now it should be up')
     } catch (e) {
-      console.info('Touched the daemon with error, anyway it should be up')
```
mipo47: Unhandled exception?

src/libraries/mysterium-client/launch-daemon/launch-daemon-process.js
```diff
@@ -65,17 +65,13 @@ class LaunchDaemonProcess implements Process {
...
+      await this._tequilapi.healthCheck()
     }
   }
 
   async stop (): Promise<void> {
     await this._tequilapi.stop()
-
-    console.info('Client Quit was successful')
```
mipo47: replace with log.info

src/renderer/dependencies/modules/bug-reporting.js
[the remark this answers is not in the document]
mipo47: From interface side stringLogger takes one string and logs it, name is ok for me.
As developer I don't need to know implementation details from class name

src/libraries/mysterium-client/launch-daemon/launch-daemon-process.js
donce: This exception is to be expected - no http server is running on `4051` port, so request results in a failure. Maybe it's possible to clarify this in code?

src/app/logger.js
donce: It has one more responsibility - it can switch logging implementations. At first, it uses `console` for logging, but after communication gets set-up, it switches to `StringLogger` (winston logger).

src/app/logging/winston-log-entry.js
donce: Good point! I've already reviewed this `switch` ~5 times, simplifying such cases would save reviewers time :)

src/app/logging/winston.js
donce: This is the same function - it was moved from `winston-log-entry.js` to here.

src/libraries/mysterium-client/launch-daemon/launch-daemon-process.js
donce: We've discussed with @shroomist not to use logging in libraries, so we can add `log.info` where this method is invoked.

src/app/logging/frontend-log-bootstrapper.js
shroomist: agree, I'm now doing exactly that with `createWinstonCachingLogger` and `createWinstonSyncComLogger`

src/app/logging/winston-log-entry.js
shroomist: ok!

src/app/logging/winston.js
shroomist: yes, it's being exported and reused

src/app/mysterion.js
shroomist: We've discussed this, majority preferred to see it explicitly `logger.setLogger(this.logger)`
here in logic, rather hidden somewhere in DI or construction

src/libraries/mysterium-client/launch-daemon/launch-daemon-process.js
shroomist: alright

src/renderer/dependencies/modules/bug-reporting.js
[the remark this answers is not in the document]
shroomist: ok

src/app/logging/backend-log-bootstrapper.js
shroomist: I think we've agreed on using `app/logger` for logging, so we should prevent devs from using `console.log`

src/app/logger.js
donce: This class is causing some confusion, because there are too many "Loggers".
Maybe calling this class `DelegateLogger` would highlight it's main purpose - to delegate logs to actual implementation?

src/libraries/mysterium-client/launch-daemon/launch-daemon-process.js
```diff
@@ -64,18 +64,17 @@ class LaunchDaemonProcess implements Process {
 
   async start (): Promise<void> {
     try {
+      // we are spawning osx Launch Daemon
       await axios.get('http://127.0.0.1:' + this._daemonPort)
```
donce: What if some other http server is using the same port? Won't `start()` method pass even though healthcheck would fail?

src/app/mysterion.js
```diff
@@ -357,6 +358,9 @@ class Mysterion {
 
     logInfo("Starting 'mysterium_client' process")
     this.process.start()
+      .then(() => { logInfo('mysterium_client start successful') })
+      .catch(() => { logException('client launch or initial healthcheck failed', new Error('mysterium_client start failed')) })
```
donce: `'client launch or initial healthcheck failed'` isn't this too technical? If we change implementation of `start()` not to use `healthcheck`, how likely are we to remember to update this comment?

src/app/logger.js
shroomist: in fact this one is the actual logger, and `winston` (anything `stringLogger` that is passed via `setLogger (logger: StringLogger)`) in our case is `delegateLogger`. so I'd rather rename `stringLogger` -> `delegateLogger`, also `logger` from DI to `delegateLogger`

src/libraries/mysterium-client/launch-daemon/launch-daemon-process.js
shroomist: true, improved with `finally { await healthcheck() }`

src/app/mysterion.js
shroomist: ok, changed the text here

src/app/logger.js
donce: Hm, why is winston logger a delagate - because it delegates work to Transports?

src/app/logging/winston-log-entry.js
```diff
@@ -1,41 +0,0 @@
...
-    case 'error':
-      return 'error'
-    case 'info':
-    case 'warn':
-    case 'debug':
-    default:
-      return 'info'
```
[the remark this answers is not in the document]
shroomist: done

shroomist:
@donce @ignasbernotas fixed

PR #236, opened 2018-07-05
MYS-121 Fix tequilapi version label and validate response DTO

donce:


src/libraries/mysterium-tequilapi/dto/node-healthcheck.js
```diff
@@ -16,20 +16,47 @@
...
+    throw new Error(errorMessage)
+  }
+
+  const version = data.version
+  if (typeof version !== 'string') {
+    throw new Error(errorMessage)
+  }
+
+  const buildInfoData = data.buildInfo
+  if (typeof buildInfoData !== 'object' || buildInfoData === null) {
```
mipo47: first you need to check for nulls, and later for type (null have no type)
if (buildInfoData === null || typeof buildInfoData !== 'object') {

src/libraries/mysterium-tequilapi/dto/node-healthcheck.js
```diff
@@ -16,20 +16,47 @@
...
+ * Tries to parse data into DTO type
+ */
+function parseHealthcheckResponse (data: mixed): NodeHealthcheckDTO {
+  const errorMessage = 'Unable to parse response'
+  if (typeof data !== 'object' || data === null) {
+    throw new Error(errorMessage)
   }
+
+  const uptime = data.uptime
+  if (typeof uptime !== 'string') {
```
mipo47: uptime is used only here, maybe replace to:
if (typeof data.uptime !== 'string') {

src/renderer/store/modules/main.js
```diff
@@ -19,57 +19,64 @@
...
+  error: ?Error,
+  showError: boolean
+}
+
+const state: State = {
   init: '',
   visual: 'head',
   navOpen: false,
-  clientBuildInfo: {},
+  clientBuildInfo: new NodeBuildInfoDTO({}),
```
mipo47: Do your really need {}? 
clientBuildInfo: new NodeBuildInfoDTO(),

src/renderer/store/modules/main.js
```diff
@@ -19,57 +19,64 @@
...
-  visual: (state: Object) => state.visual,
-  navOpen: (state: Object) => state.navOpen,
-  navVisible: (state: Object) => state.navVisible && !(state.init === type.INIT_PENDING),
-  clientBuildInfo: (state: Object) => state.clientBuildInfo,
-  errorMessage: (state: Object) => state.errorMessage,
-  showError: (state: Object) => state.showError
+  loading: (state: State) => (state.init === type.INIT_PENDING),
+  visual: (state: State) => state.visual,
+  navOpen: (state: State) => state.navOpen,
+  navVisible: (state: State) => state.navVisible && !(state.init === type.INIT_PENDING),
```
mipo47: Replace !(state.init === type.INIT_PENDING) to state.init !== type.INIT_PENDING

test/helpers/utils.js
```diff
@@ -72,16 +72,23 @@ class CallbackRecorder {
...
+  get firstArgument (): any {
+    return this.arguments[0]
+  }
+
+  get arguments (): Array<any> {
     if (!this.invoked) {
       throw new Error('CallbackRecorder: callback was not invoked, so no available argument')
     }
-    return this._argument
+    if (!this._arguments) {
```
mipo47: Calling functions without argument is not allowed?

test/unit/app/vpnInitializer.spec.js
```diff
@@ -118,17 +117,12 @@ describe('VpnInitializer', () => {
           if (args.length === 1 && args[0] === types.IDENTITY_UNLOCK_SUCCESS) {
```
mipo47: else if

test/unit/libraries/mysterium-tequilapi/client.spec.js
```diff
@@ -53,7 +54,30 @@ describe('HttpTequilapiClient', () => {
       mock.onGet('healthcheck').reply(200, response)
 
       const healthcheck = await api.healthCheck()
-      expect(healthcheck).to.deep.equal(new NodeHealthcheckDTO(response))
+      expect(healthcheck).to.deep.equal(parseHealthcheckResponse(response))
+      expect(healthcheck.version).to.eql('0.0.6')
+      expect(healthcheck.buildInfo).to.eql({
```
mipo47: Duplicated object, use response.buildInfo

test/unit/renderer/store/modules/main.spec.js
```diff
@@ -34,7 +41,8 @@ describe('mutations', () => {
     it('saves message and shows it with response error', () => {
       const state = {}
       const err = new Error('My error')
-      err.response = {
+      const errObj = (err: Object)
```
mipo47: Object -> any

test/unit/renderer/store/modules/main.spec.js
```diff
@@ -75,3 +83,38 @@ describe('mutations', () => {
...
+  describe('CLIENT_BUILD_INFO', () => {
+    it('works', async () => {
+      const client = new MainTequilapiClientMock()
+      const actions = actionsFactory(client)
+      const recorder = new CallbackRecorder()
+      const commit = recorder.getCallback()
+
+      await actions[type.CLIENT_BUILD_INFO]({ commit })
+
+      expect(recorder.arguments[0]).to.eql(type.CLIENT_BUILD_INFO)
```
mipo47: first ensure that recorder.arguments !== null and recorder.arguments.length == 2

src/renderer/store/modules/main.js
```diff
@@ -19,57 +19,64 @@
...
+import NodeBuildInfoDTO from '../../../libraries/mysterium-tequilapi/dto/node-build-info'
 
-const state = {
+type State = {
+  init: string,
+  visual: string,
+  navOpen: boolean,
+  clientBuildInfo: NodeBuildInfoDTO,
+  navVisible: boolean,
+  errorMessage: ?string,
```
mipo47: Does errorMessage and error contains same info?

src/libraries/mysterium-tequilapi/dto/node-healthcheck.js
```diff
@@ -16,20 +16,47 @@
...
+  uptime: string,
+  process: number,
+  version: string,
+  buildInfo: NodeBuildInfoDTO
+}
+
+/**
+ * Tries to parse data into DTO type
+ */
+function parseHealthcheckResponse (data: mixed): NodeHealthcheckDTO {
```
ignasbernotas: This looks to me like something that other responses could need. Maybe have a function to check if given object properties match expected types?
```
const rules = {
   uptime: 'object',
   process: 'number',
   version: 'string'
}

function checkPropertyTypes(rules, object) {
  for(let key in rules) {
    if(typeof object[key] !== rules[key]) {
      return false
    }
  }
  return true
}

if(checkPropertyTypes(rules, data)) {
throw new Error(errorMessage)
}

```

test/unit/app/vpnInitializer.spec.js
ignasbernotas: @mipo47 what for? :) General rule of thumb - if you can avoid using `else` then don't use it :)

test/helpers/utils.js
ignasbernotas: Yeah, this is strange:) good catch

src/renderer/store/modules/main.js
donce: Yes, because `clientBuildInfo` is of **class** `NodeBuildInfoDTO` type, but `{}` is of `object` type.

PR #237, opened 2018-07-10
Windows mysterium_client implementation

ignasbernotas:


test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
```diff
@@ -0,0 +1,180 @@
...
+    tequilapiPort: 4050
+  }
+
+  let system
+  describe('.needsInstallation()', () => {
+    beforeEach(() => {
+      system = new SystemMock()
+    })
+
+    it('does not need installation when all checks pass', async () => {
```
donce: `it 'returns false` would define method behaviour

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
```diff
@@ -0,0 +1,180 @@
...
+      const installer = new ServiceManagerInstaller(system, config, '/service-manager/bin/')
+      expect(await installer.needsInstallation()).to.be.eql(false)
```
donce: `.to.be.eql(false)` can be simplified to `.to.be.false`.

src/libraries/mysterium-client/service-manager/service-manager-installer.js
```diff
@@ -0,0 +1,179 @@
...
+    }
+  }
+
+  _configExists () {
+    return this._system.fileExists(this._getConfigPath())
+  }
+
+  _configChecksumMismatch () {
+    const config = md5(this._getServiceManagerConfigContents())
+    const installedConfig = md5(this._system.readFile(this._getConfigPath()))
```
donce: What do we gain from converting contents to `md5` instead of comparing contents directly?

src/main/dependencies/modules/mysterium-client.js
```diff
@@ -39,10 +44,18 @@ function bootstrap (container: Container) {
     'mysteriumClient.config',
     ['mysterionApplication.config'],
     (mysterionConfig: MysterionConfig): ClientConfig => {
+      let clientBin = path.join(mysterionConfig.contentsDirectory, 'bin', 'mysterium_client')
+      let openvpnBin = path.join(mysterionConfig.contentsDirectory, 'bin', 'openvpn')
+
+      if (os.platform() === 'win32') {
```
mipo47: Does it works for windows 64 bit?

src/main/dependencies/modules/mysterium-client.js
```diff
@@ -52,11 +65,13 @@ function bootstrap (container: Container) {
   )
   container.service(
     'mysteriumClientInstaller',
-    ['mysteriumClient.config', 'mysteriumClient.platform'],
-    (config: ClientConfig, platform: string) => {
+    ['mysterionApplication.config', 'mysteriumClient.config', 'mysteriumClient.platform'],
+    (mysterionConfig: MysterionConfig, config: ClientConfig, platform: string) => {
       switch (platform) {
```
mipo47: avoid hardcored values, put 'darwin', 'win32' to const variables

src/main/dependencies/modules/mysterium-client.js
ignasbernotas: Yep. for some reason win32 indicates both architectures:)

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
```diff
@@ -0,0 +1,180 @@
...
+      return 'SERVICE_NAME: MysteriumClient'
+    }
+
+    if (this.userExecCalledTimes === 2 && this.driverInstalled) {
+      return `123
+          'Ethernet' {F1343629-CB94-4D28-9AE4-147F9145798E}
+          asd`
+    }
+
+    return ''
```
donce: Maybe splitting mock lines from called time logic would clarify, what does actually gets returned?
```js
async userExec (command: string): Promise<string> {
  const lines = this._getExecLines()
  const line = this.userExecCalledTimes >= lines.length ? '' : lines[this.userExecCalledTimes]
  this.userExecCalledTimes++
  return line
}

_getExecLines() {
  let lines: Array<string> = []
  if (this.serviceInstalled) {
    lines.append('SERVICE_NAME: MysteriumClient')
  }
  if (this.driverInstalled) {
    lines.append(`123
            'Ethernet' {F1343629-CB94-4D28-9AE4-147F9145798E}
            asd`
    )
  }
}

```

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
```diff
@@ -0,0 +1,180 @@
...
+    // and must return different outputs
+    this.userExecCalledTimes++
+
+    if (this.userExecCalledTimes === 1 && this.serviceInstalled) {
+      return 'SERVICE_NAME: MysteriumClient'
+    }
+
+    if (this.userExecCalledTimes === 2 && this.driverInstalled) {
+      return `123
+          'Ethernet' {F1343629-CB94-4D28-9AE4-147F9145798E}
```
mipo47: Strange line alignment, suggest to align by left side:
return `123
             'Ethernet' {F1343629-CB94-4D28-9AE4-147F9145798E}
             asd`

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
ignasbernotas: you mean just 'returns false' as the description or 'returns false when all checks pass'?

src/main/dependencies/modules/mysterium-client.js
```diff
@@ -52,11 +65,13 @@ function bootstrap (container: Container) {
...
         case 'darwin':
           return new LaunchDaemonInstaller(config)
+        case 'win32':
+          return new ServiceManagerInstaller(new OSSystem(), config, path.join(mysterionConfig.contentsDirectory, 'bin'))
```
shroomist: `bin` folder contains both executables and `SERVICE_MANAGER_CONFIG`
perhaps a different folder name? `service-manager`?

src/main/dependencies/modules/mysterium-client.js
```diff
@@ -69,6 +84,8 @@ function bootstrap (container: Container) {
       switch (platform) {
         case 'darwin':
           return new LaunchDaemonProcess(tequilapiClient, LAUNCH_DAEMON_PORT, config.logDir)
+        case 'win32':
+          return new ServiceManagerProcess(tequilapiClient, config)
```
shroomist: `ServiceManagerProcess` has barely any implementation.
`MockProcess`?

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
ignasbernotas: The algorithm that checks for GUID splits the lines and in WIN/OSX environments newline character is different (\r\n vs \n) so this is a safe approach for splitting the lines.

src/libraries/mysterium-client/service-manager/service-manager-installer.js
ignasbernotas: You're right! I copied the approach from OSX:)

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
```diff
@@ -0,0 +1,180 @@
...
+      system = new SystemMock()
+    })
+
+    it('writes config file when config does not exist', async () => {
+      system.fileExistsReturnValue = false
+
+      const installer = new ServiceManagerInstaller(system, config, '/service-manager/bin/')
+      await installer.install()
+
+      expect(system.writeFileReturnValue).to.be.not.empty
```
shroomist: can't we check equality with `config` string here?

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
```diff
@@ -0,0 +1,180 @@
...
+      expect(system.sudoExecCalledTimes).to.be.eql(1)
+    })
+
+    it('installs TAP drivers when they are not installed', async () => {
+      system.driverInstalled = false
+
+      const installer = new ServiceManagerInstaller(system, config, '/service-manager/bin/')
+      await installer.install()
+
+      expect(system.userExecCalledTimes).to.be.eql(3)
```
shroomist: not so obvious how `system.userExecCalledTimes` is related to `installs TAP drivers`
could have been much better to keep track of commands being executed and match them more specific.

src/libraries/mysterium-client/system.js
```diff
@@ -0,0 +1,80 @@
...
+// @flow
+
+import { exec } from 'child_process'
+import sudo from 'sudo-prompt'
+import { promisify } from 'util'
+import fs from 'fs'
+
+const writeFile = promisify(fs.writeFile)
+
+interface System {
```
shroomist: as far as I can see, this interface takes on 2 responsibilities
1. working with fs -- reading/creating files
2. executing commands.

src/libraries/mysterium-client/system.js
```diff
@@ -0,0 +1,80 @@
...
+
+  readFile(file: string): string,
+
+  fileExists(file: string): boolean
+}
+
+class OSSystem implements System {
+  userExec (command: string): Promise<string> {
+    return new Promise(function (resolve, reject) {
+      exec(command, (error, stdout, stderr) => {
```
shroomist: using `promisify(exec)` and `async/await` could have made the code a bit easier to read

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
shroomist: does it matter here in `SystemMock` though?
anyway it's good to know this is safe for all os'es

ignasbernotas: Probably, since you should be able to run tests in Win env:)

src/main/dependencies/modules/mysterium-client.js
ignasbernotas: I think it'll be easier to pack everything into the same dir:) Less confusion and knowledge about packaging structure.

ignasbernotas: We could improve with later PRs:)

src/libraries/mysterium-client/system.js
ignasbernotas: We talked about this yesterday, promisify doesn't properly map the callback arguments.

src/libraries/mysterium-client/system.js
ignasbernotas: This can be improved, but I don't see that much value in it, especially since most of the interface can be used by OSX launch daemon implementation. Improvement commits are welcome, while "you shouldn't do that" without valid suggestions don't help:)

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
ignasbernotas: Good point! thanks!

src/main/dependencies/modules/mysterium-client.js
ignasbernotas: Will add in different PR, when logging is added.

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
```diff
@@ -0,0 +1,180 @@
...
+    })
+
+    it('returns true when config does not exits', async () => {
+      system.fileExistsReturnValue = false
+      const installer = new ServiceManagerInstaller(system, config, '/service-manager/bin/')
+      expect(await installer.needsInstallation()).to.be.eql(true)
+    })
+
+    it('returns true when config does not match existing', async () => {
+      system.readFileReturnValue = 'invalid json'
```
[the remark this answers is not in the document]
ignasbernotas: fixed

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -0,0 +1,48 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// @flow
+
+import type { ClientConfig } from '../config'
+import type { LogCallback, Process } from '../index'
+import type { TequilapiClient } from '../../mysterium-tequilapi/client'
+
+class ServiceManagerProcess implements Process {
```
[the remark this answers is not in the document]
ignasbernotas: removed config, added comment in start

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -0,0 +1,48 @@
...
+  constructor (tequilapi: TequilapiClient, config: ClientConfig) {
+    this._tequilapi = tequilapi
+    this._config = config
+  }
+
+  async start (): Promise<void> {
+  }
+
+  async stop (): Promise<void> {
+    await this._tequilapi.connectionCancel()
```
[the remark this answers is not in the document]
ignasbernotas: added comment

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -0,0 +1,48 @@
...
+import type { LogCallback, Process } from '../index'
+import type { TequilapiClient } from '../../mysterium-tequilapi/client'
+
+class ServiceManagerProcess implements Process {
+  _config: ClientConfig
+  _tequilapi: TequilapiClient
+
+  constructor (tequilapi: TequilapiClient, config: ClientConfig) {
+    this._tequilapi = tequilapi
+    this._config = config
```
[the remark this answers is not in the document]
ignasbernotas: removed:)

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
ignasbernotas: Fixed

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
ignasbernotas: fixed

src/main/dependencies/modules/mysterium-client.js
ignasbernotas: fixed

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
ignasbernotas: fixed

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
ignasbernotas: fixed

test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
ignasbernotas: fixed

src/libraries/mysterium-client/service-manager/service-manager-installer.js
ignasbernotas: Removed:)

ignasbernotas:
Fixed most comments, some things can be improved with other PRs, but this is a working version which I'd like to give out to tests ASAP:)

@donce @shroomist @mipo47

src/libraries/mysterium-client/service-manager/service-manager-installer.js
```diff
@@ -0,0 +1,178 @@
...
+  _stdoutHasGuid (stdout: string): boolean {
+    let lines = stdout.split(os.EOL)
+
+    const guidRule = new RegExp(/(\{?([A-Z0-9]{8})-(([A-Z0-9]{4}-){3})[A-Z0-9]{12}\}?)/i)
+
+    lines = lines.filter(line => {
+      return guidRule.test(line)
+    })
+
+    return lines.length > 0
```
donce: Maybe it's a good opportunity to try functional style here using `some` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) method - looks nicer and also more effective, because it returns on first `true` occurrence, something like this:
```js
const hasGuid(line => guidRule.test(line))
return lines.some(hasGuid)
```

src/libraries/mysterium-client/system.js
donce: Maybe something like `CommandExecutor` and `FileUtils` would make sense?

`CommandExecutor`
- `userExec`
- `sudoExec`

`FileUtils`
- `write`
- `read`
- `exists`

src/main/dependencies/modules/mysterium-client.js
```diff
@@ -39,10 +47,18 @@ function bootstrap (container: Container) {
     'mysteriumClient.config',
     ['mysterionApplication.config'],
     (mysterionConfig: MysterionConfig): ClientConfig => {
+      let clientBin = path.join(mysterionConfig.contentsDirectory, 'bin', 'mysterium_client')
+      let openvpnBin = path.join(mysterionConfig.contentsDirectory, 'bin', 'openvpn')
+
+      if (os.platform() === WINDOWS) {
+        clientBin += '.exe'
+        openvpnBin += '.exe'
+      }
```
donce: Maybe we can somehow keep platform-dependant logic to one place? We already have different installer implementations for different OS.

I.e. instead of `clientBin` and `openVpnBin`, we might save `binPath` and it will be up to `Installer` to know how binaries look like - with suffix or not.

PR #245, opened 2018-07-17
Show icon and code for unresolved country

mipo47:
Show icon and code for unresolved country
Send error ONCE to bugreporter when country code is not found


donce:
:-"

test/unit/renderer/components/CountrySelect.spec.js
```diff
@@ -122,7 +123,8 @@ describe('CountrySelect', () => {
       expect(wrapper.text()).to.contain('Lithuania (0x1)')
       expect(wrapper.text()).to.contain('United Kingdom (0x2)')
       expect(wrapper.text()).to.contain('N/A (0x3)')
-      expect(wrapper.text()).to.contain('N/A (0x4)')
+      expect(wrapper.text()).to.contain('N/A unknown (0x4)')
```
donce: It's not really clear that `unknown` is country code - it's confusing why there is `N/A` on one line, but `N/A unknown` on another, it seems like they are the same.
Maybe `N/A XX (0x4)` would clarify format?

src/renderer/components/CountrySelect.vue
```diff
@@ -96,7 +97,11 @@ export default {
       return getCountryLabel(country)
     },
     imagePath (code) {
-      if (!code) {
+      if (!countryFound(code)) {
+        if (this.unresolvedCountryList.indexOf(code) < 0) {
+          this.unresolvedCountryList.push(code)
+          this.bugReporter.captureInfoMessage('Country not found, code: ' + code)
```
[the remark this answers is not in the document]
mipo47: in most cases yes, not not here

src/app/countries/index.js
```diff
@@ -29,9 +30,12 @@ type Country = {
 
 function getCountryLabel (country: Country, maxNameLength: ?number = null, maxIdentityLength: ?number = 9) {
   const identity = limitedLengthString(country.id, maxIdentityLength)
-  const name = limitedLengthString(country.name, maxNameLength)
+  let title = limitedLengthString(country.name, maxNameLength)
+  if (title === COUNTRY_NAME_UNRESOLVED && country.code) {
```
donce: ?? How does this work - title has max length 2, so unresolved title will be `'N/`, which is not equal to `'N/A'`

mipo47: magic of if

test/unit/renderer/components/CountrySelect.spec.js
```diff
@@ -122,7 +121,11 @@ describe('CountrySelect', () => {
       expect(wrapper.text()).to.contain('Lithuania (0x1)')
       expect(wrapper.text()).to.contain('United Kingdom (0x2)')
       expect(wrapper.text()).to.contain('N/A (0x3)')
-      expect(wrapper.text()).to.contain('N/A (0x4)')
+      expect(wrapper.text()).to.contain('N/A un.. (0x4)')
+
+      expect(wrapper.vm.unresolvedCountryList).to.contain('unknown')
```
donce: ?? Why do we need to assert `wrapper.vm.unresolvedCountryList` - to make sure that it's saved and the second invocation will not trigger bug reporter?

Maybe we can test behaviour directly - i.e. triggering unresolved country second time and making sure that bug reporter mock isn't triggered?

test/helpers/bug-reporter-mock.js
```diff
@@ -20,20 +20,42 @@
...
+  context: any
+}
+
+type StringCapture = {
+  message: string,
+  context: any
+}
+
 class BugReporterMock implements BugReporter {
+  identity: IdentityDTO
```
donce: ?? This public property doesn't seem to be used anywhere, maybe we can avoid unused public properties?

test/helpers/bug-reporter-mock.js
```diff
@@ -20,20 +20,42 @@
...
   captureErrorMessage (_message: string, _context?: any): void {
+    this.errorMessages.push({ message: _message, context: _context })
   }
 
   captureInfoMessage (_message: string, _context?: any): void {
+    this.infoMessages.push({ message: _message, context: _context })
   }
 
   captureErrorException (_err: Error, _context?: any): void {
+    this.errorExceptions.push({ error: _err, context: _context })
```
donce: !! The similar thing is done in `VpnLoaderBugReporter`, which extends `BugReporterMock`. Maybe we don't need `VpnLoaderBugReporter` anymore and we can use `BugReporterMock` directly instead?

test/helpers/bug-reporter-mock.js
mipo47: Tried to make universal mock for future, now only one property is used

test/unit/renderer/components/CountrySelect.spec.js
mipo47: We assert it because it's easy, would be better to write separate test for wrapper.vm.unresolvedCountryList?

test/helpers/bug-reporter-mock.js
mipo47: ok, VpnLoaderBugReporter removed

test/unit/renderer/components/CountrySelect.spec.js
donce: As I see it, `unresolvedCountryList` in internal state - it's part of another feature of "invoking bug reporter ONLY the first time". I.e. `if (this.unresolvedCountryList.indexOf(code) < 0) {` this conditional is not tested at all, although that's the purpose of `unresolvedCountryList`.

test/unit/renderer/pages/VpnLoader.spec.js
```diff
@@ -160,7 +152,8 @@ describe('VpnLoader', () => {
     })
 
     it('reports error', () => {
-      const error = bugReporter.errorException
+      expect(bugReporter.errorExceptions.length).to.eql(1)
+      const error = bugReporter.errorExceptions[0].error
       if (!error) {
```
donce: This conditional is not required anymore - `bugReporter.errorExceptions[0].error` is not of **maybe** type anymore.

PR #251, opened 2018-07-25
MYS-143 Windows build

donce:


util_scripts/download-win-binaries.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
tadovas: Why not reuse download-binaries.sh? The only difference is just .exe at the end

donce: There are more differences than just .exe:
- uzing `unzip`instead of `tar`
- no need to `chown +x`
- need to implement temporary fix, because openvpn is missing .exe

In code it would be easy to separate common parts, but as for bash, I'm not sure whether it's worth extracting common place, because there are quite a few uncommon places.

PR #253, opened 2018-07-26
MYS-128 Recover windows service

mipo47:
Recover works in following cases:
- Tequilapi stopped using stop call
- MysteriumClient process killed
- MysteriumClient process crashes
- MysteriumClient service is stopped
- MysteriumClient freezes and doesn't responds


src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -17,18 +17,77 @@
...
+      const serviceManagerPath = path.join(this._serviceManagerDir, SERVICE_MANAGER_BIN)
+      const operation = state === SERVICE_STATE.RUNNING ? 'restart' : 'start'
+      const command = `${serviceManagerPath} --do=${operation}`
+      logger.info('Running command', command)
+      const serviceInfo = await this._system.sudoExec(command)
+      state = this._parseServiceState(serviceInfo)
+
+      // wait until service will be started
+      while (state !== SERVICE_STATE.RUNNING) {
+        state = await this._serviceState()
```
shroomist: uh-oh, infinite loop on `serviceState()`? 0.o
do you think it's better to throw after some number of retries?

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -43,6 +102,34 @@ class ServiceManagerProcess implements Process {
 
   onLog (level: string, cb: LogCallback): void {
   }
+
+  async _serviceState (): Promise<ServiceState> {
```
shroomist: this method name is not descriptive enough -- its better to name functions as verbs.
eg `getServiceState`

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -17,21 +17,89 @@
...
+      const command = `${serviceManagerPath} --do=${operation}`
+      logger.info('Running command', command)
+      const serviceInfo = await this._system.sudoExec(command)
+      state = this._parseServiceState(serviceInfo)
+
+      const now = () => (new Date()).getTime()
+      const startTime = now()
+
+      // wait until service will be started
+      while (state !== SERVICE_STATE.RUNNING && now() - startTime < this._serviceInitTime) {
```
shroomist: this `try{}` block grows big and complex and difficult to grasp. especially now when we define timing functions.
I suggest moving some of the `const` values out of this `try{}` block and leaving inside just the code that can throw

src/main/dependencies/modules/mysterium-client.js
```diff
@@ -118,13 +118,13 @@ function bootstrap (container: Container) {
 
   container.service(
     'mysteriumClientProcess',
-    ['tequilapiClient', 'mysteriumClient.config', 'mysteriumClient.logSubscriber', 'mysteriumClient.platform'],
-    (tequilapiClient: TequilapiClient, config: ClientConfig, logSubscriber: ClientLogSubscriber, platform: string) => {
+    ['tequilapiClient', 'mysteriumClient.config', 'mysteriumClient.logSubscriber', 'mysteriumClient.platform', 'mysterionApplication.config'],
```
shroomist: I think it's good to keep `mysterion` and `mysterium-client` configs separate.
so better to add `serviceManagerBinPath` to `mysteriumClient.config`

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -17,21 +17,89 @@
...
+      logger.info(`Service state: [${state}]`)
+      if (state === SERVICE_STATE.START_PENDING) {
+        return
+      }
+      if (this._startingFirstTime && state === SERVICE_STATE.RUNNING) {
+        return
+      }
+      const serviceManagerPath = path.join(this._serviceManagerDir, SERVICE_MANAGER_BIN)
+      const operation = state === SERVICE_STATE.RUNNING ? 'restart' : 'start'
+      const command = `${serviceManagerPath} --do=${operation}`
```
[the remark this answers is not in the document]
mipo47: in this class it's not duplicated and will be replaced with native commands later

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -17,21 +17,91 @@
...
   }
 
   async start (): Promise<void> {
-    // mysterium_client is started automatically by WIN service
+    const state = await this._getServiceState()
+    if (state === SERVICE_STATE.RUNNING) {
+      return
+    }
+
+    await this.repair()
```
ignasbernotas: would be clearer if start would call `_doStart` directly rather than `repair`

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -17,21 +17,91 @@
...
+  }
+
+  async repair (): Promise<void> {
+    await this._doStart()
+  }
+
+  async _doStart (state: ?ServiceState = null): Promise<void> {
```
ignasbernotas: I can't find a place where _doStart is called with an argument

mipo47: fixed

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -17,21 +17,66 @@
...
+type ServiceState = $Values<typeof SERVICE_STATE>
 
 class ServiceManagerProcess implements Process {
   _tequilapi: TequilapiClient
   _logs: ClientLogSubscriber
+  _serviceManagerDir: string
+  _serviceInitTime: number
+  _system: System
+  _monitoring: Monitoring
+  _fixIsRunning: boolean
```
donce: `_fixIsRunning` should be `?boolean`, or default value `false` should be specified.

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -17,21 +17,66 @@
...
 
-  constructor (tequilapi: TequilapiClient, logs: ClientLogSubscriber) {
+  constructor (
+    tequilapi: TequilapiClient,
+    logs: ClientLogSubscriber,
+    serviceManagerDir: string,
+    system: System,
+    monitoring: Monitoring,
+    serviceInitTime: number = SERVICE_INIT_TIME) {
```
donce: Unused parameter - it's always skipped.

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -47,6 +92,90 @@ class ServiceManagerProcess implements Process {
...
     this._logs.onLog(level, cb)
   }
+
+  async _fixService (state: ?ServiceState = null): Promise<void> {
+    if (this._fixIsRunning) {
+      return
+    }
+    this._fixIsRunning = true
+    try {
+      state = state || await this._getServiceState()
```
donce: We can move this out to `repair` - that way `state` will be always passed in, which simplifies flow.

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -47,6 +92,90 @@ class ServiceManagerProcess implements Process {
...
+      logger.info(`Service state: [${state}]`)
+      if (state === SERVICE_STATE.START_PENDING) {
+        return
+      } else if (state === SERVICE_STATE.UNKNOWN) {
+        throw new Error('Cannot start non-installed service')
+      }
+      const serviceManagerPath = path.join(this._serviceManagerDir, SERVICE_MANAGER_BIN)
+      const operation = state === SERVICE_STATE.RUNNING ? 'restart' : 'start'
```
donce: What about `STOP_PENDING` case - do we want to `restart` in that case?

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -47,6 +92,90 @@ class ServiceManagerProcess implements Process {
...
+          resolve()
+        }
+      }
+      this._monitoring.onStatus(statusCallback)
+    })
+
+    if (statusCallback) {
+      this._monitoring.removeOnStatus(statusCallback)
+    }
+  }
```
donce: This looks quite generic - we already have very similar implementation `onFirstEventOrTimeout`.
Main difference - that event doesn't unsubscribe the callback.

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -47,6 +92,90 @@ class ServiceManagerProcess implements Process {
...
+
+  async _getServiceState (): Promise<ServiceState> {
+    let stdout
+    try {
+      stdout = await this._system.userExec(`sc.exe query "${SERVICE_NAME}"`)
```
donce: This command is duplicated in 12 places, can we DRY it up?

src/libraries/mysterium-client/service-manager/service-manager-process.js
mipo47: also it doesn't checks isRunning

src/libraries/mysterium-client/service-manager/service-manager-installer.js
```diff
@@ -33,11 +34,13 @@ class ServiceManagerInstaller implements Installer {
   _system: System
   _config: ClientConfig
   _serviceManagerDir: string
+  _serviceManager: ServiceManager
 
   constructor (system: System, config: ClientConfig, serviceManagerDir: string) {
     this._system = system
     this._config = config
     this._serviceManagerDir = serviceManagerDir
+    this._serviceManager = new ServiceManager(path.join(serviceManagerDir, SERVICE_MANAGER_BIN), system)
```
ignasbernotas: !! Object instantiation should be done outside of the class, instead this should be injected as one of the arguments. This will make the object less heavy, plus it'll allow mocking it if needed.

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -17,21 +17,58 @@
...
+    tequilapi: TequilapiClient,
+    logs: ClientLogSubscriber,
+    serviceManagerDir: string,
+    system: System,
+    monitoring: Monitoring) {
     this._tequilapi = tequilapi
     this._logs = logs
+    this._serviceManagerDir = serviceManagerDir
+    this._serviceManager = new ServiceManager(path.join(this._serviceManagerDir, SERVICE_MANAGER_BIN), system)
+    this._system = system
```
ignasbernotas: !! Same as for the installer. Object should be instantiated outside of the class scope. Once that is done, _system property is no longer needed in this class.

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -0,0 +1,105 @@
...
+    }
+    return ServiceManager._parseServiceState(stdout)
+  }
+
+  async _execAndGetState (commandName: string) {
+    const result = await this._system.sudoExec(`${this._path} --do=${commandName}`)
+    return ServiceManager._parseServiceState(result)
+  }
+
+  static _parseServiceState (serviceInfo: string): ServiceState {
```
ignasbernotas: Does it have to be static if it's only used in this class?

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -0,0 +1,105 @@
...
+    const isInstalled: boolean = serviceInfo.indexOf(`SERVICE_NAME: ${SERVICE_NAME}`) > -1
+    if (!isInstalled) {
+      return SERVICE_STATE.UNKNOWN
+    }
+
+    let start = serviceInfo.indexOf('STATE')
+    if (start < 0) {
+      return SERVICE_STATE.UNKNOWN
+    }
```
ignasbernotas: ?? It's not a big deal, but not having newlines between logical structures makes it harder to read code. Do you prefer it this way?

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -47,6 +84,48 @@ class ServiceManagerProcess implements Process {
...
+    this._fixIsRunning = true
+    try {
+      state = state || await this._serviceManager.getServiceState()
+      logger.info(`Service state: [${state}]`)
+      if (state === SERVICE_STATE.START_PENDING) {
+        return
+      } else if (state === SERVICE_STATE.UNKNOWN) {
+        throw new Error('Cannot start non-installed service')
+      }
+      if (state === SERVICE_STATE.RUNNING) {
```
ignasbernotas: ?? It's not a big deal, but not having newlines between logical structures makes it harder to read code. Do you prefer it this way?

src/libraries/mysterium-client/service-manager/service-manager.js
mipo47: yes, it's private static because used only in this class and doesn't depends on class fields

src/libraries/mysterium-client/service-manager/service-manager-process.js
shroomist: looks like a case for a `switch(state)`

src/libraries/mysterium-client/service-manager/service-manager.js
shroomist: somehow we are not using many statics, and seem to prefer simple functions instead. please consider.

mipo47: why?

src/libraries/mysterium-client/service-manager/service-manager-process.js
mipo47: agree, fixed

src/libraries/mysterium-client/service-manager/service-manager.js
shroomist: in this particular case you are exporting `ServiceManager` and you want `_parseServiceState` to be private.
with a simple function definition this method is not leaked out unless you export it.

mipo47: thanks, fixed

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -17,21 +17,55 @@
...
-    // mysterium_client is started automatically by WIN service
+    const state = await this._serviceManager.getServiceState()
+    if (state === SERVICE_STATE.RUNNING) {
+      return
+    }
+
+    await this._fixService(state)
+  }
+
+  async repair (): Promise<void> {
```
shroomist: where is this repair being used? sorry, couldn't find it

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -47,6 +81,49 @@ class ServiceManagerProcess implements Process {
   onLog (level: string, cb: LogCallback): void {
     this._logs.onLog(level, cb)
   }
+
+  async _fixService (state: ?ServiceState = null): Promise<void> {
```
shroomist: ?? not sure why we need both public `repair()` and private `_fixService()`

ignasbernotas: Repair is an interface method without arguments, I assume because of that :)

test/helpers/system-mock.js
```diff
@@ -0,0 +1,92 @@
...
+  setMockFile (path: string, content: string): void,
+  unsetMockFile (path: string): void,
+
+  setMockCommand (command: string, response: string): void,
+  unsetMockCommand (command: string): void
+}
+
+export default class SystemMock implements System, SystemMockManager {
+  writeFileReturnValue: ?string = null
+  userExecCalledCommands: Array<string> = []
```
shroomist: userExecCalledCommands:string[]

src/libraries/mysterium-client/service-manager/service-manager-process.js
shroomist: got it

src/libraries/mysterium-client/service-manager/service-manager-process.js
mipo47: in mysterion when health check not passed:
```
    this.monitoring.onStatus((status) => {
      if (status === false) {
        logInfo("Starting 'mysterium_client' process, because it's currently down")
        this.process.repair()
      }
    })
```

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -47,6 +81,49 @@ class ServiceManagerProcess implements Process {
...
+    } finally {
+      this._fixIsRunning = false
+    }
+  }
+
+  async _waitForHealthCheck (): Promise<void> {
+    let statusCallback: ?StatusCallback
+    await new Promise((resolve, reject) => {
+      const rejectTimer = setTimeout(() => reject(new Error('Unable to start service')), SERVICE_INIT_TIME)
+      statusCallback = (isRunning: boolean) => {
```
shroomist: can be more readable if we name this `resolveAndClearTimer`

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -0,0 +1,115 @@
...
+}
+type ServiceState = $Values<typeof SERVICE_STATE>
+
+export { SERVICE_STATE }
+export type { ServiceState }
+
+const parseServiceState = (serviceInfo: string): ServiceState => {
+  const isInstalled: boolean = serviceInfo.indexOf(`SERVICE_NAME: ${SERVICE_NAME}`) > -1
+  if (!isInstalled) {
+    return SERVICE_STATE.UNKNOWN
```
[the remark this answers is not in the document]
mipo47: this state is also used when we can't parse state

src/app/mysterion.js
```diff
@@ -411,22 +411,37 @@ class Mysterion {
...
     })
 
     logInfo("Starting 'mysterium_client' monitoring")
     this.monitoring.start()
   }
 
+  async _repairProcess () {
+    try {
+      await this.process.repair()
+    } catch (e) {
```
shroomist: !! we should also send this error via `bugReporter`

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -47,6 +81,49 @@ class ServiceManagerProcess implements Process {
...
+    }
+    this._fixIsRunning = true
+    try {
+      state = state || await this._serviceManager.getServiceState()
+      logger.info(`Service state: [${state}]`)
+      switch (state) {
+        case SERVICE_STATE.START_PENDING:
+          return
+        case SERVICE_STATE.UNKNOWN:
+          throw new Error('Cannot start non-installed service')
```
shroomist: `'Cannot start non-installed or broken service'`

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -17,21 +17,55 @@
...
+import ServiceManager, { SERVICE_STATE } from './service-manager'
+import type { ServiceState } from './service-manager'
+
+/***
+ * Time in milliseconds required to fully activate Mysterium client
+ * @type {number}
+ */
+const SERVICE_INIT_TIME = 8000
 
 class ServiceManagerProcess implements Process {
```
shroomist: I think we can throw away `Managers` because they dont add value. :) @ignasbernotas?

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -0,0 +1,123 @@
...
+    this._path = serviceManagerPath
+    this._system = system
+  }
+
+  get directory (): string {
+    return path.dirname(this._path)
+  }
+
+  async install (): Promise<string> {
+    return this._sudoExec(`${this._path} --do=install`)
```
shroomist: !! `return await this._sudoExec(${this._path} --do=install`
also I'm not sure if we want to return the output of `sudoExec`.
so perhaps parse output and do error transformations here

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -0,0 +1,123 @@
...
+  async _execAndGetState (commandName: string): Promise<ServiceState> {
+    const result = await this._sudoExec(`${this._path} --do=${commandName}`)
+    return parseServiceState(result)
+  }
+
+  async _sudoExec (command: string): Promise<string> {
+    try {
+      return await this._system.sudoExec(command)
+    } catch (e) {
+      throw new Error(`Unable to start "${SERVICE_NAME}" service. ${e}`)
```
shroomist: depending on a `command` this error message might be wrong

src/libraries/mysterium-client/service-manager/service-manager.js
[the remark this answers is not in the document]
shroomist: I suggest we introduce `PARSING_ERROR` or `ERROR` and `NOT_INSTALLED`
we can use `serviceInfo.includes(SERVICE_NAME)` to mane sure it's installed

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -0,0 +1,123 @@
...
+    return SERVICE_STATE.UNKNOWN
+  }
+
+  const length = serviceInfo.indexOf('\r', start) - start - 1
+  if (length < 0) {
+    return SERVICE_STATE.UNKNOWN
+  }
+
+  const state = serviceInfo.substr(start, length)
+  if (Object.values(SERVICE_STATE).indexOf(state) < 0) {
```
shroomist: it is confusing for me a bit, so the enum values map directly to windows service output?

src/libraries/mysterium-client/monitoring.js
```diff
@@ -69,9 +89,14 @@ class Monitoring {
...
-    this._notifySubscribers(isRunning)
-
-    this._timer = setTimeout(() => this._healthCheckLoop(), healthCheckInterval)
+    try {
+      this._notifySubscribers(isRunning)
+    } catch (e) {
+      e.message = 'Bad subscriber added to Monitoring: ' + e.message
+      throw e
+    } finally {
+      this._timer = setTimeout(() => this._healthCheckLoop(), healthCheckInterval)
```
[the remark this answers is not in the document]
mipo47: context will be missed in this case, to keep right context we need to write:
this._timer = setTimeout(this._healthCheckLoop.bind(this), healthCheckInterval)

src/libraries/mysterium-client/service-manager/service-manager.js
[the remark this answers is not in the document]
mipo47: service name in windows response doesn't means that service is installed
(ex: unable to find MysteriumClient)
That's why I try to parse state to ensure that windows response is valid

src/libraries/mysterium-client/monitoring.js
```diff
@@ -33,23 +33,43 @@ class Monitoring {
   _subscribersStatus: Array<StatusCallback> = []
   _subscribersUp: Array<UpCallback> = []
   _subscribersDown: Array<DownCallback> = []
+  _isStarted: boolean = false
```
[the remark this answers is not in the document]
mipo47: similar field is already defined "lastIsRunning" for another purpose, so I decide to pick different name

src/libraries/mysterium-client/service-manager/service-manager.js
mipo47: It's validation: state should be equal to one of SERVICE_STATE values. Flow can't check enum types at runtime

src/app/mysterion.js
mipo47: fixed

PR #255, opened 2018-07-27
Test coverage

donce:
`> TEST_COVERAGE=true yarn unit`
```
=============================== Coverage summary ===============================
Statements   : 53.89% ( 928/1722 )
Branches     : 57.75% ( 231/400 )
Functions    : 45.81% ( 268/585 )
Lines        : 54.03% ( 918/1699 )
================================================================================
```
screen shot 2018-07-27 at 15 00 01

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+test/unit/coverage/**/*.js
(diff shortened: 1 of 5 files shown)


shroomist:
we should add `test/unit/coverage` to lint ignore, because now after running coverage the linter is going to constantly fail for that folder

PR #258, opened 2018-07-30
[hotfix] Disabled system log tail when file doesn't exist

ignasbernotas:
This was causing an error in windows since file didn't exist.

### DIFF
--- a/src/libraries/mysterium-client/client-log-subscriber.js
+++ b/src/libraries/mysterium-client/client-log-subscriber.js
@@ -16,7 +16,7 @@
  */
 
 // @flow
-
+import fs from 'fs'
 import Subscriber from '../subscriber'
 import logLevels from './log-levels'
 import type { LogCallback } from './index'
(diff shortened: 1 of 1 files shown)


src/libraries/mysterium-client/client-log-subscriber.js
```diff
@@ -65,25 +81,18 @@ class ClientLogSubscriber {
       notifyOnErrorSubscribers(prependWithCurrentTime(prependWithSpace(data)))
     })
 
+    // not all OSs have a system file
+    if (!fs.existsSync(this._systemFilePath)) {
```
shroomist: this one was the problem, right?
how about dumping service-manager logs here instead?

ignasbernotas: Service manager itself writes to stdout log:)

shroomist: tell me more about it :)

donce: So if windows doesn't have this file, we should not pass `systemFilePath = '/var/log/system.log'` value - error seems to be in `mysteriumClient.config`.
As for cases when we are expecting system.log to be present, but it's not - we shouldn't ignore this error, we should report this to Sentry to be aware whether this happens often.

ignasbernotas: fixed

src/main/dependencies/modules/mysterium-client.js
```diff
@@ -50,14 +51,16 @@ function bootstrap (container: Container) {
...
+    (platform: string, mysterionConfig: MysterionConfig): ClientConfig => {
       let clientBin = path.join(mysterionConfig.contentsDirectory, 'bin', 'mysterium_client')
       let openvpnBin = path.join(mysterionConfig.contentsDirectory, 'bin', 'openvpn')
+      let systemLogPath = '/var/log/system.log'
 
-      if (os.platform() === WINDOWS) {
+      if (platform === WINDOWS) {
         clientBin += '.exe'
         openvpnBin += '.exe'
+        systemLogPath = null
```
shroomist: it is not necessary to nullify, as long as we check this way
`if (!systemFilePath) {`
both `undefined` and `null` evaluate to `false`

PR #265, opened 2018-08-06
MYS-161 API endpoint 'identity/{id}/registration' -> identityRegistration

mipo47:


src/libraries/mysterium-tequilapi/dto/identity-registration.js
```diff
@@ -0,0 +1,55 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// @flow
+
+class PublicKeyDTO {
+  part1: string
+  part2: string
+
+  constructor (data: Object) {
```
ignasbernotas: Object is a loose type, since these classes are exported, I'd like a custom type instead of Object

src/libraries/mysterium-tequilapi/dto/identity-registration.js
```diff
@@ -0,0 +1,55 @@
...
+    this.part2 = data.Part2
+  }
+}
+
+class SignatureDTO {
+  r: string
+  s: string
+  v: string
+
+  constructor (data: Object) {
```
ignasbernotas: Object is a loose type, since these classes are exported, I'd like a custom type instead of Object. Also, is it possible to map r/s/v to human readable names? I have no idea what these are for.

src/libraries/mysterium-tequilapi/dto/identity-registration.js
```diff
@@ -0,0 +1,55 @@
...
+    this.v = data.V
+  }
+}
+
+class IdentityRegistrationDTO {
+  registered: string
+  publicKey: PublicKeyDTO
+  signature: SignatureDTO
+
+  constructor (data: Object) {
```
ignasbernotas: Object is a loose type, since these classes are exported, I'd like a custom type instead of Object

src/libraries/mysterium-tequilapi/dto/identity-registration.js
mipo47: but it' the way all DTO are implemented, should I introduce new way?

src/libraries/mysterium-tequilapi/dto/identity-registration.js
mipo47: it's parts of Elliptic Curve signature, if we want rename them - better to do it in API side first

src/libraries/mysterium-tequilapi/dto/identity-registration.js
ignasbernotas: Nah, nevermind:)

PR #266, opened 2018-08-06
[MYS-148] Feature toggle

ignasbernotas:
https://mysteriumnetwork.atlassian.net/browse/MYS-148

### DIFF
--- a/.electron-vue/features.js
+++ b/.electron-vue/features.js
@@ -1,0 +1,28 @@
+'use strict'
+
+const fs = require('fs')
+const path = require('path')
+
+const getFeatures = function (file) {
+  if (!fs.existsSync(file)) {
+    return null
+  }
+
+  try {
+    return JSON.parse(fs.readFileSync(file))
+  } catch (e) {
+  }
+
+  return null
+}
+
+const productionFeatures = getFeatures(path.join(__dirname, '../', 'features.prod.json'))
+const localFeatures = getFeatures(path.join(__dirname, '../', 'features.json'))
+
+if (!productionFeatures) {
+  throw new Error('Failed to read features.prod.json. This file is required for building.')
+}
+
+const features = Object.assign(productionFeatures, localFeatures)
+
+module.exports = features
--- a/src/app/feature-toggle.js
+++ b/src/app/feature-toggle.js
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2017 The "MysteriumNetwork/mysterion" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// @flow
+
+type Features = {
+  payments: boolean
+}
+
+class FeatureToggle {
+  _features: Features
+
+  constructor (features: Features) {
+    this._features = features
+  }
+
+  paymentsAreEnabled () {
+    return this._features.payments
+  }
+}
+
+export default FeatureToggle
+export type { Features }
--- a/src/main/dependencies/modules/application.js
+++ b/src/main/dependencies/modules/application.js
@@ -18,6 +18,7 @@
 // @flow
 import { app, BrowserWindow } from 'electron'
 import type { Container } from '../../../app/di'
+import FeatureToggle from '../../../app/feature-toggle'
 import Mysterion from '../../../app/mysterion'
 import type { MysterionConfig } from '../../../app/mysterionConfig'
 import path from 'path'
@@ -68,6 +69,15 @@ function bootstrap (container: Container) {
   )
 
   container.service(
+    'featureToggle',
+    [],
+    () => {
+      declare var FEATURES: any
+      return new FeatureToggle(FEATURES)
+    }
+  )
+
+  container.service(
     'mysterionApplication',
     [
       'mysterionApplication.config',
@@ -83,7 +93,8 @@ function bootstrap (container: Container) {
       'mysteriumProcessLogCache',
       'bugReporterMetrics',
       'userSettingsStore',
-      'disconnectNotification'
+      'disconnectNotification',
+      'featureToggle'
     ],
     (
       mysterionConfig: MysterionConfig,
@@ -99,7 +110,8 @@ function bootstrap (container: Container) {
       mysteriumProcessLogCache,
       bugReporterMetrics,
       userSettingsStore,
-      disconnectNotification
+      disconnectNotification,
+      featureToggle
     ) => {
       return new Mysterion({
         config: mysterionConfig,
@@ -118,7 +130,8 @@ function bootstrap (container: Container) {
         frontendLogCache,
         mysteriumProcessLogCache,
         userSettingsStore,
-        disconnectNotification
+        disconnectNotification,
+        featureToggle
       })
     }
   )
--- a/src/renderer/dependencies/modules/application.js
+++ b/src/renderer/dependencies/modules/application.js
@@ -19,6 +19,7 @@
 import { Container } from '../../../app/di'
 import RendererCommunication from '../../../app/communication/renderer-communication'
 import RendererIpc from '../../../app/communication/ipc/renderer-ipc'
+import FeatureToggle from '../../../app/feature-toggle'
 import ElkCollector from '../../../app/statistics/elk-collector'
 import AggregatingCollector from '../../../app/statistics/aggregating-collector'
 import NullCollector from '../../../app/statistics/null-collector'
@@ -36,12 +37,11 @@ function bootstrap (container: Container) {
 
   container.service(
     'rendererCommunication',
-    ['bugReporterMetrics'],
-    (bugReporterMetrics) => {
+    [],
+    () => {
       const ipc = new RendererIpc()
       const messageBus = new IpcMessageBus(ipc)
-      const communication = new RendererCommunication(messageBus)
-      return communication
+      return new RendererCommunication(messageBus)
     }
   )
 
@@ -52,6 +52,7 @@ function bootstrap (container: Container) {
       version: mysterionReleaseID
     }
   )
+
   container.service(
     'statsEventFactory',
     ['statsApplicationInfo'],
@@ -59,6 +60,7 @@ function bootstrap (container: Container) {
       return createEventFactory(applicationInfo)
     }
   )
+
   container.service(
     'statsCollector',
     [],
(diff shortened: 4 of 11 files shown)


.electron-vue/features.js
```diff
@@ -0,0 +1,28 @@
...
+  try {
+    return JSON.parse(fs.readFileSync(file))
+  } catch (e) {
+  }
+
+  return null
+}
+
+const productionFeatures = getFeatures(path.join(__dirname, '../', 'features.prod.json'))
+const localFeatures = getFeatures(path.join(__dirname, '../', 'features.json'))
```
mipo47: path.join(__dirname, '../' used twice, would be nice to have variable FEATURES_FOLDER = path.join(__dirname, '../')

ignasbernotas: Cool, I'll DRY it up a bit, thanks:)

src/main/dependencies/modules/application.js
```diff
@@ -67,6 +68,15 @@ function bootstrap (container: Container) {
     }
   )
 
+  container.service(
+    'featureToggle',
+    [],
+    () => {
+      declare var FEATURES: any
```
shroomist: you could be more specific with type here eg `declare var FEATURES: { [string]: boolean }`

src/app/feature-toggle.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+class FeatureToggle {
+  _features: Features
+
+  constructor (features: Features) {
+    this._features = features
+  }
+
+  paymentsAreEnabled () {
```
shroomist: we could make this method a bit more generic
`isEnabled(feature: string): boolean { return !!this._features[feature] }`
`!!` <- casts anything into right boolean

ignasbernotas: I don't like that we have to know the feature name string in order to check if that feature is enabled, having a method seems cleaner:)

.electron-vue/features.js
```diff
@@ -0,0 +1,31 @@
+'use strict'
```
donce: ?? Are you sure this file can't be in Flow? I've tried changing it, everything seems to be working fine..

.electron-vue/features.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  return null
+}
+
+const productionFeatures = getFeaturesFromFile('features.prod.json')
+const localFeatures = getFeaturesFromFile('features.json')
+
+if (!productionFeatures) {
+  throw new Error('Failed to read features.prod.json. This file is required for building.')
```
donce: Can we extract `features.prod.json` as a `const` so that we shouldn't repeat it in this message?

src/renderer/dependencies/modules/application.js
```diff
@@ -71,6 +73,17 @@ function bootstrap (container: Container) {
...
     }
   )
+
+  container.service(
+    'featureToggle',
+    [],
+    () => {
+      declare var FEATURES: { [string]: boolean }
+
+      return new FeatureToggle(FEATURES)
```
donce: !! In practice, `FEATURES` can be `undefined` here, but `FeatureToggle` assumes it's defined. Maybe it's worth specifying `FeatureToggle` constructor param as optional, `constructor (features: ?Features)` ?

.electron-vue/features.js
```diff
@@ -0,0 +1,31 @@
...
+  try {
+    return JSON.parse(fs.readFileSync(path))
+  } catch (e) {
+  }
+
+  return null
+}
+
+const productionFeatures = getFeaturesFromFile('features.prod.json')
+const localFeatures = getFeaturesFromFile('features.json')
```
donce: Executing code when importing file is dangerous - isn't it safer to export `readFeatures()` function instead? As a bonus, this would allow us to test this function.

src/app/feature-toggle.js
```diff
@@ -0,0 +1,37 @@
...
+    return this._features.payments
+  }
+}
+
+export default FeatureToggle
+export type { Features }
```
donce: !! This export doesn't seem to be used anywhere.

src/app/feature-toggle.js
```diff
@@ -0,0 +1,37 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// @flow
+
+type Features = {
+  payments: boolean
+}
+
+class FeatureToggle {
```
donce: This class is pretty simple, so tests doesn't seem very valuable, but still, maybe it's worth testing since tests should be simple as well?
I.e. what happens when `features` is `undefined`, which can actually happen if our webpack scripts fail?

src/app/features/feature-toggle.js
```diff
@@ -0,0 +1,54 @@
...
+    return this._getFeatureState('payments', false)
+  }
+
+  _getFeatureState (key: string, defaultValue: boolean): boolean {
+    if (typeof this._features !== 'object') {
+      return defaultValue
+    }
+
+    const features = this._features
+    if (features == null) {
```
donce: !! This is being checked twice: line 36 already does that, but it uses `this._features` instead of `feature`.

src/app/features/feature-toggle.js
```diff
@@ -0,0 +1,54 @@
...
+      return defaultValue
+    }
+
+    if (typeof features[key] === 'undefined') {
```
donce: From https://flow.org/en/docs/types/maybe/:

> You can simplify the two checks against null and undefined using a single != null check which will do both.

So this is the same as `if (features[key] == null)`

src/app/features/feature-toggle.js
```diff
@@ -0,0 +1,54 @@
...
+      return defaultValue
+    }
```
mipo47:
```
     if (features == null) {
       return defaultValue
     }

     if (typeof features[key] === 'undefined') {
       return defaultValue
     }
```
to
```
if (!features) {
  return defaultValue
}
```

src/app/features/read-features.js
```diff
@@ -0,0 +1,33 @@
...
+    return null
+  }
+
+  try {
+    return JSON.parse(fs.readFileSync(path).toString())
+  } catch (e) {
+  }
+
+  return null
+}
```
donce: Instead of returning `null`, I would throw error - that way, we can explicitly ignore **local** features in `features.js` and add a better error message for **production** features - i.e. 
```js
try {
  const localFeatures = readFeatures(getFeatureFilePath(LOCAL_FILE))
} catch (err) {
  // local file is optional
}

try {
  const productionFeatures = readFeatures(getFeatureFilePath(PRODUCTION_FILE))
} catch (err) {
  throw new Error(`Failed to read required file: ${PRODUCTION_FILE} : err.toString()`)
}
```

src/main/dependencies/modules/application.js
```diff
@@ -68,6 +70,16 @@ function bootstrap (container: Container) {
     }
   )
 
+  container.service(
+    'featureToggle',
+    [],
+    () => {
+      declare var FEATURES: Features
```
donce: `: ?Features`, because we can't be sure it's present.

src/app/features/read-features.js
```diff
@@ -0,0 +1,33 @@
...
+
+function readFeatures (path) {
+  if (!fs.existsSync(path)) {
+    return null
+  }
+
+  try {
+    return JSON.parse(fs.readFileSync(path).toString())
+  } catch (e) {
+  }
```
mipo47: log error / send to Sentry

test/unit/app/features/feature-toggle.spec.js
```diff
@@ -0,0 +1,40 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// @flow
+
+import FeatureToggle from '../../../../src/app/features/feature-toggle'
+import { describe, expect, it } from '../../../helpers/dependencies'
+
+describe('.paymentsAreEnabled', () => {
```
donce: !! To keep it consistent and clear what's being tested, we should add class description: `describe('FeatureToggle')`

shroomist:
please fix comments above, and don't forget to remove `needs feedback` when you're still working on the branch.

.electron-vue/features.js
shroomist: ?? still no flow here?

src/app/features/read-features.js
```diff
@@ -0,0 +1,33 @@
...
+  if (!fs.existsSync(path)) {
+    throw new Error(`Feature file doesn't exist in "${path}"`)
+  }
+
+  try {
+    return JSON.parse(fs.readFileSync(path).toString())
+  } catch (e) {
+  }
+
+  throw new Error('Unable to parse "' + path + '" feature file')
```
shroomist: this should be thrown at L:27

src/main/dependencies/modules/application.js
```diff
@@ -68,6 +70,16 @@ function bootstrap (container: Container) {
     }
   )
 
+  container.service(
+    'featureToggle',
+    [],
+    () => {
+      declare var FEATURES: ?Features
```
shroomist: declarations of globals should be on top of the file or in separate file

src/renderer/dependencies/modules/application.js
```diff
@@ -31,12 +33,21 @@ function bootstrap (container: Container) {
...
-      return communication
+      return new RendererCommunication(messageBus)
+    }
+  )
+
+  container.service(
+    'featureToggle',
+    [],
+    () => {
+      declare var FEATURES: ?Features
```
shroomist: duplicate declaration -> move to external file

.electron-vue/features.js
shroomist: I suggest moving this file out of `.electron-vue` to the root of our project or to `src/` folder

.electron-vue/features.js
```diff
@@ -0,0 +1,28 @@
...
+let localFeatures = {}
+
+try {
+  productionFeatures = readFeatures(getFeatureFilePath(PRODUCTION_FILE))
+} catch (e) {
+  throw new Error(`Failed to read ${PRODUCTION_FILE}. This file is required for building.`)
+}
+
+try {
+  localFeatures = readFeatures(getFeatureFilePath(LOCAL_FILE))
```
shroomist: no need to parse yourself, you can simply
`import features from ('./features.json')`
or
`var features = require('./features.json')`

donce: Local file is optional, so we can't require it.

src/renderer/dependencies/modules/application.js
donce: Actually, the whole block of `'featureToggle'` is duplicated - we can extract it whole.

ignasbernotas:
@shroomist fixed comments that I felt needed changes. IMO the rest is fine, the feature works. Flow cannot be used here, because this code is ran by nodejs directly. @donce @mipo47 please review

.electron-vue/features.js
```diff
@@ -0,0 +1,23 @@
...
+const getFeatureFilePath = (fileName) => path.join(__dirname, '../', fileName)
+
+let localFeatures = {}
+
+const productionFeatures = readFeatures(getFeatureFilePath(PRODUCTION_FILE))
+
+try {
+  localFeatures = readFeatures(getFeatureFilePath(LOCAL_FILE))
+} catch (e) {
+
```
donce: I would add a comment why this catch is false, i.e. `// local feature file is optional`

PR #269, opened 2018-08-09
Feature/mys 174 reinstall service

mipo47:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mysterion",
-  "version": "0.1.0",
+  "version": "0.2.1",
```
donce: !! This should not be updated here.

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -82,19 +91,29 @@ export default class ServiceManager {
   }
 
   async install (): Promise<string> {
-    return this._sudoExec(`${this._path} --do=install && ${this._path} --do=start`)
+    return this._sudoExec(`${escapePath(this._path)} --do=install && ${escapePath(this._path)} --do=start`)
```
shroomist: maybe we can `escapePath` in constructor?

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -68,6 +68,11 @@ const parseServiceState = (serviceInfo: string): ServiceState => {
   return (state: ServiceState)
 }
 
+const needReinstall = (e): boolean => {
```
shroomist: !! type for `e` missing

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -82,42 +87,76 @@ export default class ServiceManager {
...
+  async _sudoExec (...commands: Command[]): Promise<string> {
+    try {
+      logger.info('Execute sudo', JSON.stringify(commands))
+      return await this._system.sudoExec(...commands)
+    } catch (e) {
+      throw new Error(`Unable to execute [${JSON.stringify(commands)}]. ${e}`)
+    }
+  }
+
+  _createCommand (commandName: string): Command {
```
shroomist: `_createCommand` seems to belong in `'../system` module

src/libraries/mysterium-client/system.js
```diff
@@ -98,7 +108,7 @@ class OSSystem implements System {
     return new Promise(function (resolve, reject) {
       sudo.exec(command, { name: 'Mysterion' }, (error, stdout, stderr) => {
         if (error) {
-          reject(error)
+          reject(wrapError(error))
```
shroomist: ?? in which cases do we get an error object here of a different type than `Error`?

src/libraries/mysterium-client/system.js
```diff
@@ -48,6 +48,16 @@ const stringifyCommands = (commands: Command[]) => {
...
 }
 
+const wrapError = (e: Object): Error => {
+  if (e instanceof Error) {
+    return e
+  }
+  const error = new Error(e.toString())
+  const errorObj = (error: Object)
+  errorObj.original = e
+  throw error
```
shroomist: ?? `wrapError` should return Error obj here, but throws it, is it intentional?
what's happening with stack trace in this case?
should we pull out this function in `libraries` and test it accordingly?

src/libraries/mysterium-client/system.js
```diff
@@ -85,7 +95,7 @@ class OSSystem implements System {
     return new Promise(function (resolve, reject) {
       exec(command, (error, stdout, stderr) => {
         if (error) {
-          reject(error)
+          reject(wrapError(error))
```
shroomist: do we reject the promise properly here if `wrapError`throws or do we just throw up?

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -68,6 +70,16 @@ const parseServiceState = (serviceInfo: string): ServiceState => {
...
 }
 
+const needReinstall = (e: Error | Object): boolean => {
+  let message: string
+  if (e instanceof Error) {
+    message = e.message
+  } else {
+    message = e.toString()
+  }
+  return message.indexOf('Command failed') >= 0
```
donce: !! `message.includes('Command failed')`

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -111,27 +132,38 @@ export default class ServiceManager {
...
+    return {
+      path: this._path,
+      args: [`--do=${operation}`]
+    }
   }
 
-  async _execAndGetState (commandName: string): Promise<ServiceState> {
-    const result = await this._execCommands(commandName)
-    return parseServiceState(result)
+  async _execAndGetState (operation: ServiceManagerOperation, reinstallOnError: boolean = false): Promise<ServiceState> {
```
shroomist: this now does 2 things execs operation, and reinstalls on error.

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -68,6 +70,16 @@ const parseServiceState = (serviceInfo: string): ServiceState => {
   return (state: ServiceState)
 }
 
+const needReinstall = (e: Error | Object): boolean => {
+  let message: string
+  if (e instanceof Error) {
+    message = e.message
+  } else {
+    message = e.toString()
```
donce: ?? Do we need to handle `else` case, can't we just `return false` if it's not an `Error` as expected?

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -82,19 +94,28 @@ export default class ServiceManager {
   }
 
   async install (): Promise<string> {
-    return this._execCommands('install', 'start')
+    return this._execOperations('install', 'start')
+  }
+
+  async reinstall (): Promise<string> {
+    let commands = ['uninstall', 'install', 'start']
```
donce: `const`

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -111,27 +132,38 @@ export default class ServiceManager {
...
+      const result = await this._execOperations(operation)
+      state = parseServiceState(result)
+    } catch (e) {
+      if (reinstallOnError && needReinstall(e)) {
+        await this.reinstall()
+      } else {
+        throw e
+      }
+    }
+    return state
```
donce: Very hard to understand this code, i.e. why `state = SERVICE_STATE.START_PENDING` is default.
`SERVICE_STATE.START_PENDING` comes from reinstallation part, so it can be there:
```js
try {
      const result = await this._execOperations(operation)
      return parseServiceState(result)
    } catch (e) {
      if (reinstallOnError && needReinstall(e)) {
        await this.reinstall()
        return SERVICE_STATE.START_PENDING
      } else {
        throw e
      }
    }
```

src/libraries/mysterium-client/system.js
donce: ?? Why do we `return` in one case, but `throw` in another one?

src/libraries/mysterium-client/system.js
```diff
@@ -48,6 +48,16 @@ const stringifyCommands = (commands: Command[]) => {
     .join(COMMANDS_SEPARATOR)
 }
 
+const wrapError = (e: Object): Error => {
```
donce: Nice function, looks like we can re-use it in the future. 👍

donce: If you pass `Error` object to `wrapError`, it doesn't actually wrap anything - maybe `ensureErrorObject` or `ensureError` would be more accurate?

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -111,27 +127,39 @@ export default class ServiceManager {
...
-  async _sudoExec (...commands: Command[]): Promise<string> {
+  async _execAndGetState (operation: ServiceManagerOperation, reinstallOnError: boolean = false): Promise<ServiceState> {
+    let state = SERVICE_STATE.UNKNOWN
     try {
-      return await this._system.sudoExec(...commands)
+      const result = await this._execOperations(operation)
+      state = parseServiceState(result)
     } catch (e) {
-      throw new Error(`Unable to execute [${JSON.stringify(commands)}]. ${e}`)
+      if (reinstallOnError && needReinstall(e)) {
```
shroomist: ??
it seems that we always `reinstallOnError` unless we do `stop` operation.
perhaps we could simplify this method signature by doing something along the lines
```
async _execAndGetState (operation: ServiceManagerOperation) {
  let reinstallOnError = true
  if (operation == 'stop') reinstallOnError = false;
...
```

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -70,14 +70,9 @@ const parseServiceState = (serviceInfo: string): ServiceState => {
...
 
-const needReinstall = (e: Error | Object): boolean => {
-  let message: string
-  if (e instanceof Error) {
-    message = e.message
-  } else {
-    message = e.toString()
-  }
-  return message.indexOf('Command failed') >= 0
+const needReinstall = (e: Error): boolean => {
```
donce: Since this function is handling any `Object` param, not just `Error`, we should tell that in signature:
`(errorObj: Object): boolean`

src/libraries/mysterium-client/service-manager/service-manager.js
```diff
@@ -111,27 +127,39 @@ export default class ServiceManager {
...
+  _createCommandFromOperation (operation: ServiceManagerOperation): Command {
+    return {
+      path: this._path,
+      args: [`--do=${operation}`]
+    }
   }
 
-  async _sudoExec (...commands: Command[]): Promise<string> {
+  async _execAndGetState (operation: ServiceManagerOperation, reinstallOnError: boolean = false): Promise<ServiceState> {
+    let state = SERVICE_STATE.UNKNOWN
```
donce: !! This statement is redundant - all flow branches override this value.

src/libraries/mysterium-client/service-manager/service-manager.js
mipo47: Function expects only error type, because we all errors are wrapped in exec

src/libraries/mysterium-client/service-manager/service-manager.js
mipo47: state will be unknown when reinstall failes

PR #271, opened 2018-08-10
Service manager working dir change

ignasbernotas:
Fixes issue with the latest mysterium_client where `config` dir isn't found in the working dir.


test/unit/libraries/mysterium-client/service-manager/service-manager-installer.spec.js
```diff
@@ -25,11 +25,15 @@ import type { SystemMockManager } from '../../../../helpers/system-mock'
 import type { System } from '../../../../../src/libraries/mysterium-client/system'
 import ServiceManager from '../../../../../src/libraries/mysterium-client/service-manager/service-manager'
 
+const SERVICE_MANAGER_DIR = '/service-manager/bin'
+const SERVICE_MANAGER_PATH = SERVICE_MANAGER_DIR + '/servicemanager.exe'
+const CONFIG_FILE = SERVICE_MANAGER_DIR + '/servicemanager.json'
```
donce: Using `path.join` would avoid issues with `/` characters - we could just skip them

PR #278, opened 2018-08-17
[MYS-182] [MYS-141] Send 'platform' metric to ELK with 'startup' event

donce:
Adding 'startup' event with `platform` field.
screen shot 2018-08-17 at 08 59 58
screen shot 2018-08-17 at 09 29 48

This also closes [MYS-141]:
screen shot 2018-08-17 at 09 42 15


src/app/statistics/collector-event-sender.js
```diff
@@ -0,0 +1,41 @@
...
+
+// @flow
+
+import type { EventCollector, EventFactory } from './events'
+import type { EventSender } from './event-sender'
+
+/**
+ * Event sender, which builds events using `EventFactory` and sends them to `EventCollector`.
+ */
+class CollectorEventSender implements EventSender {
```
shroomist: how about `CollectorEventSender` -> `EventSenderToCollector`
or even shorter `EventToCollector` because only one class method `eventToCollector.send(name, ctx)`

src/app/statistics/event-sender-to-collector.js
```diff
@@ -0,0 +1,41 @@
...
+
+// @flow
+
+import type { EventCollector, EventFactory } from './events'
+import type { EventSender } from './event-sender'
+
+/**
+ * Event sender, which builds events using `EventFactory` and sends them to `EventCollector`.
+ */
+class EventSenderToCollector implements EventSender {
```
mipo47: Very tiny responsibility of class, I recommend to move it away,
just add send method to EventCollector

src/app/mysterion.js
```diff
@@ -116,6 +119,7 @@ class Mysterion {
     this.mysteriumProcessLogCache = params.mysteriumProcessLogCache
     this.userSettingsStore = params.userSettingsStore
     this.disconnectNotification = params.disconnectNotification
+    this._startupEventTracker = params.startupEventTracker
```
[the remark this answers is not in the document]
donce: Unified naming - renamed remaining fields with `_` prefix, because all those fields are private.

src/app/statistics/event-sender-to-collector.js
donce: Don't think that's so simple - `EventCollector` in interface with 4 classes behind it, so we would introduce a lot of duplication separated across different classes.

PR #280, opened 2018-08-20
MYS 146 show identity registration status

mipo47:
screen shot 2018-08-21 at 13 46 01

Please update Mysterium Client after this PR


src/renderer/components/identity-registration.vue
```diff
@@ -0,0 +1,44 @@
...
+  name: 'IdentityRegistration',
+  dependencies: ['rendererCommunication'],
+  data () {
+    return {
+      registered: null
+    }
+  },
+  mounted () {
+    this.rendererCommunication.onRegistrationUpdate(r => {
+      this.registered = r
```
ignasbernotas: !! use full variable names r => registered, had to go to the method definition to check its type

mipo47: Changed argument name, but unable to set type, because flow is not supported in VUE files

test/unit/app/data-fetchers/registration-fetcher.spec.js
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    describe('.start', () => {
+      it('triggers subscriber callbacks', async () => {
+        let counter = 0
+
+        fetcher.onFetchedRegistration(() => counter++)
+        fetcher.start(IDENTITY_ID)
+
+        await tickWithDelay(1000)
```
shroomist: ?? interval is 1001 and we tick only 1000 here.
event still happens.
can you explain a bit what's going on with this here?

mipo47: first time is called in start(), 
committed update to make it more obvious

PR #281, opened 2018-08-20
[MYS-180] Change network errors capturing level to "INFO"

donce:
- All network errors are captured as `info` exceptions in Sentry - this is done with `BugReporterAdapter`.
- Since all network errors are being captured in adapter, no need to capture it at feature-level - only unknown errors are being captured there.
- Did some refactoring on specs I've touched as well, because there is quite a lot of mess there and it's hard to understand and extend them :/


src/libraries/mysterium-tequilapi/client-factory.js
```diff
@@ -21,18 +21,24 @@ import axios from 'axios'
...
 import { TIMEOUT_DEFAULT } from './timeouts'
 import HttpTequilapiClient from './client'
+import BugReporterAdapter from './adapters/bug-reporter-adapter'
+import type { BugReporter } from '../../app/bug-reporting/interface'
 
 const TEQUILAPI_URL = 'http://127.0.0.1:4050'
 
-function tequilapiClientFactory (baseUrl: string = TEQUILAPI_URL, defaultTimeout: number = TIMEOUT_DEFAULT): HttpTequilapiClient {
+function tequilapiClientFactory (
+  bugReporter: BugReporter,
```
shroomist: tequilapi should not know about bugReporter. ;)

src/libraries/mysterium-tequilapi/client-factory.js
```diff
@@ -21,18 +21,24 @@ import axios from 'axios'
...
+  defaultTimeout: number = TIMEOUT_DEFAULT): HttpTequilapiClient {
   const axiosInstance = axios.create({
     baseURL: baseUrl,
     headers: {
       'Cache-Control': 'no-cache, no-store'
     }
   })
   const axiosAdapter = new AxiosAdapter(axiosInstance, defaultTimeout)
-  return new HttpTequilapiClient(axiosAdapter)
+  const adapter = new BugReporterAdapter(axiosAdapter, bugReporter)
```
shroomist: I'd rather be putting this adapter somewhere in `src/app`
or at least in a separate `Factory`

src/libraries/mysterium-tequilapi/tequilapi-client-factory.js
```diff
@@ -21,18 +21,35 @@ import axios from 'axios'
...
+class TequilapiClientFactory {
+  _baseUrl: string
+  _defaultTimeout: number
+
+  constructor (baseUrl: string = TEQUILAPI_URL, defaultTimeout: number = TIMEOUT_DEFAULT) {
+    this._baseUrl = baseUrl
+    this._defaultTimeout = defaultTimeout
+  }
+
+  build (): TequilapiClient {
```
shroomist: ?? I wonder why don't we build in constructor?

test/helpers/adapters/MockHttpAdapter.js
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// @flow
+
+import type { HttpInterface, HttpQueryParams } from '../../../src/libraries/mysterium-tequilapi/adapters/interface'
+
+class MockHttpAdapter implements HttpInterface {
```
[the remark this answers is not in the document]
donce: Testing with `MockHttpAdapter` seems simpler, especially asserting `timeout` and request body/params - that gets complicated at `axios-mock-adapter` level.

Such specs already exist in `axios-adapter.spec.js` - so we are already covering most of http level tests.
For this adapter, I only care that it invokes internal adapter with correct arguments, I don't care whether internal adapter makes http requests or not.

PR #282, opened 2018-08-20
Run copyright header check in yarn ci

ignasbernotas:
Builds fail when this check fails, and I keep forget to run manually it so I think it could be included in the `yarn ci` command:)

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "test:integration:start-server": "karma start test/integration/karma.conf.js",
     "unit:start-server": "karma start test/unit/karma.conf.js",
     "unit:run": "karma run test/unit/karma.conf.js",
-    "ci": "yarn flow && yarn lint && yarn unit && yarn test:integration && yarn test:coverage",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -26,7 +26,7 @@
     "test:integration:start-server": "karma start test/integration/karma.conf.js",
     "unit:start-server": "karma start test/unit/karma.conf.js",
     "unit:run": "karma run test/unit/karma.conf.js",
-    "ci": "yarn flow && yarn lint && yarn unit && yarn test:integration && yarn test:coverage",
+    "ci": "util_scripts/check-headers.sh && yarn flow && yarn lint && yarn unit && yarn test:integration && yarn test:coverage",
```
donce: I would keep `yarn flow` the first one, because it usually runs instantly and gives very quick feedback :)

ignasbernotas: Header check is quicker than flow (it needs to boot up some server or something..) so I've placed it at the beginning

donce: Maybe, but headers fail very rarely when WebStorm is configured to auto-insert headers. Showing Flow errors faster helps :)

PR #283, opened 2018-08-22
Feature/mys 183 os ver stats

shroomist:
in addition to platform, we're now sending osName and os.release, along with identity.

### DIFF
--- a/src/app/communication/main-message-bus-communication.js
+++ b/src/app/communication/main-message-bus-communication.js
@@ -123,6 +123,14 @@ class MainMessageBusCommunication implements MainCommunication, MapSyncCommunica
     this._on(messages.CURRENT_IDENTITY_CHANGED, callback)
   }
 
+  onCurrentIdentityChangeOnce (callback: (CurrentIdentityChangeDTO) => void) {
+    let cb = (currentIdentityChange: CurrentIdentityChangeDTO) => {
+      callback(currentIdentityChange)
+      this._messageBus.removeCallback(messages.CURRENT_IDENTITY_CHANGED, cb)
+    }
+    this._on(messages.CURRENT_IDENTITY_CHANGED, cb)
+  }
+
   onProposalUpdateRequest (callback: () => void) {
     this._on(messages.PROPOSALS_UPDATE, callback)
   }
--- a/src/app/mysterion.js
+++ b/src/app/mysterion.js
@@ -195,8 +195,12 @@ class Mysterion {
     const ipc = new MainIpc(send, this._bugReporter.captureErrorException)
     this._messageBus = new IpcMessageBus(ipc)
     this._communication = new MainMessageBusCommunication(this._messageBus)
-    this._communication.onCurrentIdentityChange((identityChange: CurrentIdentityChangeDTO) => {
+
+    this._communication.onCurrentIdentityChangeOnce((identityChange: CurrentIdentityChangeDTO) => {
       this._startupEventTracker.sendEvent(identityChange.id)
+    })
+
+    this._communication.onCurrentIdentityChange((identityChange: CurrentIdentityChangeDTO) => {
       const identity = new IdentityDTO({ id: identityChange.id })
       this._bugReporter.setUser(identity)
 
--- a/src/app/mysterion.js
+++ b/src/app/mysterion.js
@@ -177,8 +177,6 @@ class Mysterion {
   }
 
   async bootstrap () {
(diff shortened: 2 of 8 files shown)


src/app/mysterion.js
```diff
@@ -198,6 +196,7 @@ class Mysterion {
     this._messageBus = new IpcMessageBus(ipc)
     this._communication = new MainMessageBusCommunication(this._messageBus)
     this._communication.onCurrentIdentityChange((identityChange: CurrentIdentityChangeDTO) => {
+      this._startupEventTracker.sendEvent(identityChange.id)
```
donce: !! This event will be triggered after each `VpnInitializer` cycle - this means that after sleep or random `mysterium_client` error, we'll receive another `startup` event - we should ensure this event is sent only once.

src/app/statistics/startup-event-tracker.js
```diff
@@ -32,8 +33,13 @@ class StartupEventTracker {
...
   }
 
-  async sendEvent () {
-    const context = { platform: os.platform() }
+  async sendEvent (uid: string) {
+    const context = {
+      platform: os.platform(),
+      osName: osName(os.platform(), os.release()),
+      release: os.release(),
+      identity: uid
```
donce: Why calling it `uid` and not `identity`, as it's called elsewhere?

src/app/statistics/startup-event-tracker.js
```diff
@@ -32,8 +33,13 @@ class StartupEventTracker {
     this._eventSender = eventSender
   }
 
-  async sendEvent () {
-    const context = { platform: os.platform() }
+  async sendEvent (uid: string) {
+    const context = {
+      platform: os.platform(),
+      osName: osName(os.platform(), os.release()),
+      release: os.release(),
```
donce: ?? Name seems confusing - it looks more like application release, not OS release. `osRelease` / `platformRelease`?

src/app/statistics/startup-event-tracker.js
donce: !! We should extend specs to test that new values are being sent.

test/unit/app/communication/main-message-bus-communication.spec.js
```diff
@@ -0,0 +1,44 @@
...
+  let mainCommunication
+  const callback = () => {}
+
+  beforeEach(() => {
+    messageBus = new SubscribableMessageBus()
+    mainCommunication = new MainMessageBusCommunication(messageBus)
+  })
+
+  describe('.onCurrentIdentityChangeOnce', () => {
+    it('fires once only', () => {
```
ignasbernotas: ?? seems like you're not testing multiple event triggers, even though the test says so, or am I missing something?:)

src/app/mysterion.js
```diff
@@ -197,6 +195,11 @@ class Mysterion {
     const ipc = new MainIpc(send, this._bugReporter.captureErrorException)
     this._messageBus = new IpcMessageBus(ipc)
     this._communication = new MainMessageBusCommunication(this._messageBus)
+
```
ignasbernotas: Mysterion class is growing bigger and bigger, maybe it's time to think about moving out new parts out of it somehow? Just a thought:)

test/unit/app/communication/main-message-bus-communication.spec.js
shroomist: yes, this test makes sure that no callbacks are remaining once we fire the related event.

src/app/mysterion.js
shroomist: we've registered a tech debt tasks related to that.
we might start deconstructing as soon as we're done with the sprint tickets (pretty possible this week)

test/unit/app/communication/main-message-bus-communication.spec.js
ignasbernotas: Ah, got it, cheers

src/app/communication/main-message-bus-communication.js
```diff
@@ -123,6 +123,14 @@ class MainMessageBusCommunication implements MainCommunication, MapSyncCommunica
     this._on(messages.CURRENT_IDENTITY_CHANGED, callback)
   }
 
+  onCurrentIdentityChangeOnce (callback: (CurrentIdentityChangeDTO) => void) {
+    let cb = (currentIdentityChange: CurrentIdentityChangeDTO) => {
```
donce: `const`

PR #284, opened 2018-08-22
Feature/mys 147 instructions for registering

mipo47:
screen shot 2018-08-23 at 15 49 02

Shows instructions and generate link for payment.
Don't forget to use latest Mysterium Client


src/app/logger.js
```diff
@@ -19,6 +19,10 @@
 
 import type { StringLogger } from './logging/string-logger'
 
+const stringifyArgs = (data: Array<any>) => {
```
ignasbernotas: !! I'd like to see a test for this, seems straightforward, but not sure about the end result of the function.

src/renderer/components/identity-registration.vue
```diff
@@ -16,28 +16,85 @@
...
     }
   },
   mounted () {
-    this.rendererCommunication.onRegistrationUpdate(isRegistered => {
-      this.registered = isRegistered
+    this.rendererCommunication.onRegistrationUpdate(registration => {
+      this.registered = registration.registered
+
+      const { publicKey, signature } = registration
+      this.paymentLink = `http://walletx.mysterium.network/` +
```
ignasbernotas: !! I think this link building should be moved out of the component, or at least tested:)

src/app/logger.js
shroomist: hmm. unrelated changes, somebody wants tests for it..
perhaps pull this out for a separate PR, so that we don't block this one?

mipo47: Unit test added

src/libraries/mysterium-tequilapi/dto/identity-registration.js
```diff
@@ -17,6 +17,13 @@
 
 // @flow
 
+const getPaymentLink = (registration: IdentityRegistrationDTO): string => {
+  const { publicKey, signature } = registration
+  return `http://walletx.mysterium.network/` +
```
shroomist: I still highly suggest we move this in some sort of config in DI

test/unit/app/logger.spec.js
```diff
@@ -0,0 +1,119 @@
...
+    })
+
+    it('logs simple string', () => {
+      const str = 'THIS_IS_STRING'
+      logger.info(str)
+      expect(stringLogger.infoText).to.be.eql(`"${str}"`)
+    })
+
+    it('logs number as string', () => {
+      const str = 123.45
```
ignasbernotas: var name shouldn't be str if it's not a string

src/app/communication/main-message-bus-communication.js
```diff
@@ -79,8 +80,8 @@ class MainMessageBusCommunication implements MainCommunication, MapSyncCommunica
     this._send(messages.COUNTRY_UPDATE, countries)
   }
 
-  sendRegistration (registered: boolean) {
-    this._send(messages.IDENTITY_REGISTERED, registered)
+  sendRegistration (registration: IdentityRegistrationDTO) {
+    this._send(messages.IDENTITY_REGISTERED, registration)
```
donce: Sending `IDENTITY_REGISTERED` message with `registration` dto is very confusing - it seems like this message is being sent only when identity is registered.

`messages.IDENTITY_REGISTRATION` seems better, and more consistent with method names.

src/libraries/mysterium-tequilapi/dto/identity-registration.js
```diff
@@ -49,7 +56,12 @@ class IdentityRegistrationDTO {
     this.publicKey = new PublicKeyDTO(data.publicKey || {})
     this.signature = new SignatureDTO(data.signature || {})
   }
+
+  getPaymentLink (paymentBaseUrl: string): string {
```
donce: !! This is not used anywhere, only in spec.
Maybe we can use it in `identity-registration.vue`?

src/renderer/components/identity-registration.vue
```diff
@@ -16,28 +16,89 @@
...
+        <li>2. Claim MYST and ETH test tokens</li>
+        <li>3. Allow Mysterium SmartContract to reserve MYST tokens</li>
+        <li>4. Register your ID by clicking on “Pay & Register For ID”</li>
+        <li>5. Wait for few minutes until the payment is processed</li>
+      </ul>
+      <div
+        class="btn"
+        v-if="registration"
+        @click="openPaymentsUrl()">
+        Copy Wallet Address
```
donce: !! Button name here and in instructions differ.

src/renderer/components/identity-registration.vue
```diff
@@ -16,28 +16,89 @@
...
+      <div class="identity-text">ID</div>
+      <div
+        class="identity-tooltip"
+        v-if="!registered">Please activate your ID</div>
+    </div>
+
+    <div
+      class="app__nav nav is-open"
+      id="registration-instructions"
+      v-if="showInstructions">
```
donce: ?? This shows instructions instantly, without sliding animation as in left menu. Can we have animation here as well?

src/renderer/components/identity-registration.vue
```diff
@@ -16,28 +16,89 @@
...
+        return null
+      }
+      return this.registration.registered
     }
   },
   mounted () {
-    this.rendererCommunication.onRegistrationUpdate(isRegistered => {
-      this.registered = isRegistered
+    this.rendererCommunication.onRegistrationUpdate(registration => {
+      this.registration = registration
```
donce: Making a UI test to assert that this UI changes once registration update is received might be useful.

mipo47: Added

test/unit/app/logger.spec.js
```diff
@@ -78,6 +78,16 @@ describe('Logger', () => {
       logger.info(arg1, arg2, arg3)
       expect(stringLogger.infoText).to.be.eql('"abc" 3.14159265 {"x":1,"y":2}')
     })
+
+    it('logs object with circular reference', () => {
+      const object = {
+        str: 'any string',
+        child: {}
+      }
+      object['child'] = object
```
donce: Why `object['child']` and not simply `object.child`?

test/unit/app/logger.spec.js
```diff
@@ -115,5 +125,12 @@ describe('Logger', () => {
       logger.error(arg1, arg2, arg3)
       expect(stringLogger.errorText).to.be.eql('"a b c" 2.7182818 {"x":"1","y":2}')
     })
+
+    it('logs custom Error', () => {
+      const error = new Error('ERROR_MESSAGE')
+      error.stack = 'MOCKED_STACK'
```
donce: Stack is not used.

src/app/logger.js
```diff
@@ -20,7 +20,25 @@
...
+  return data.map(d => {
+    if (d instanceof Error) {
+      return d.toString()
+    }
+
+    // http://stackoverflow.com/questions/11616630/json-stringify-avoid-typeerror-converting-circular-structure-to-json/11616993#11616993
+    const cache = []
+    return JSON.stringify(d, (key, value) => {
+      if (typeof value === 'object' && value !== null) {
+        if (cache.indexOf(value) !== -1) {
```
donce: `cache` should be `Set` - that's what they are for: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

src/app/logger.js
```diff
@@ -20,7 +20,25 @@
...
+          // Circular reference found, discard key
+          return
+        }
+        // Store value in our collection
+        cache.push(value)
+      }
+      return value
+    })
```
donce: This lamba is huge, and it's not that obvious the structure of this function.
Extracting function might help:
```js
return data
  .map(stringifyArg)
  .join(' ')
```

test/unit/renderer/components/identity-registration.spec.js
```diff
@@ -0,0 +1,81 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// @flow
+import { beforeEach, describe, expect, it } from '../../../helpers/dependencies'
+import DIContainer from '../../../../src/app/di/vue-container'
+import IdentityRegistration from '@/components/identity-registration'
+import { createLocalVue, mount } from '@vue/test-utils'
+
+class FakeRendererCommunication {
```
donce: !! This is incomplete mock it's type is different from what UI actually expects.
To have full implementation and avoid duplication, we can re-use real `RendererCommunication` with `DirectMessageBus` - this allows to send any messages in the tests.

test/unit/renderer/components/identity-registration.spec.js
```diff
@@ -0,0 +1,81 @@
...
+    rendererCommunication = new FakeRendererCommunication()
+    dependencies.constant('rendererCommunication', rendererCommunication)
+    dependencies.constant('paymentBaseUrl', '')
+    vue = mount(IdentityRegistration, {
+      localVue: vm
+    })
+  })
+
+  describe('renders', () => {
+    it('no ID icon until registration state comes from communication', () => {
```
donce: !! `it` should specify behaviour, `it` + description should make a sentence, i.e. `it('renders no ID icon...`. That way, maybe we don't need `describe('renders')`?

test/unit/renderer/components/identity-registration.spec.js
```diff
@@ -0,0 +1,81 @@
...
+
+    it('ID icon when identity becomes registered', () => {
+      rendererCommunication.callback({ registered: true })
+      expect(vue.findAll('.identity-registration')).to.have.lengthOf(1)
+      expect(vue.findAll('.identity-registered')).to.have.lengthOf(1)
+      expect(vue.findAll('.identity-unregistered')).to.have.lengthOf(0)
+    })
+
+    it('ID icon when identity becomes registered', () => {
+      rendererCommunication.callback({ registered: false })
```
donce: !! spec descprition says 'when .. becomes registered', but implementation is different.

PR #286, opened 2018-08-23
[MYS-83] Tequilapi error

donce:
- Fix `Error` extending using `babel-plugin-transform-builtin-extend` plugin.
- Add `TequilapiError` as an example for easier error handling


test/unit/renderer/store/modules/connection.spec.js
```diff
@@ -344,6 +351,18 @@ describe('connection', () => {
         const committed = await executeAction(type.CONNECTION_IP)
         expect(committed).to.eql([])
       })
+
+      it('captures unknown errors', async () => {
+        fakeTequilapi.setIpFail()
+        await executeAction(type.CONNECTION_IP)
+        expect(bugReporterMock.errorExceptions.length).to.eql(1)
```
mipo47: to.have.lengthOf

src/libraries/mysterium-tequilapi/tequilapi-error.js
```diff
@@ -0,0 +1,72 @@
...
+    super(message)
+
+    this._originalError = originalError
+  }
+
+  get code (): ?string {
+    return this._originalError.code || null
+  }
+
+  isTimeoutError (): boolean {
```
shroomist: !! I suggest making this class instances serialisation resistant.
the thing we've noticed, is that when serialising for message-bus we are loosing the methods, perhaps the inheritance tree as well.
how about making changing these methods into members variables, assigning value on construction?

src/libraries/mysterium-tequilapi/adapters/axios-adapter.js
```diff
@@ -74,10 +74,7 @@ async function decorateResponse (promise: Promise<Object>, path: string): Promis
   try {
     response = await promise
   } catch (err) {
-    markErrorAsHttp(err)
-    err.name = 'Tequilapi error'
-    err.message = `${err.message} (path="${path}")`
-    throw err
+    throw new TequilapiError(err, path)
```
mipo47: !! Also put request data to TequilaError, sometimes it can be important to reproduce error

src/libraries/mysterium-tequilapi/tequilapi-error.js
```diff
@@ -0,0 +1,72 @@
...
+
+  constructor (originalError: Error, path: string) {
+    const message = `${originalError.message} (path="${path}")`
+    super(message)
+
+    this._originalError = originalError
+  }
+
+  get code (): ?string {
+    return this._originalError.code || null
```
shroomist: no need to `|| null` the value.
we can leave it `undefined`

src/libraries/mysterium-tequilapi/tequilapi-error.js
```diff
@@ -0,0 +1,72 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// @flow
+
+class TequilapiError extends Error {
+  name: string = 'TequilapiError'
```
mipo47: !! Can't find usage of this field, only in test

src/libraries/mysterium-tequilapi/tequilapi-error.js
donce: Fixed.

test/unit/libraries/mysterium-tequilapi/adapters/axios-adapter.spec.js
```diff
@@ -71,18 +68,16 @@ describe('TequilapiClient AxiosAdapter', () => {
...
   it('returns timeout error', async () => {
     mock.onGet('test-url').timeout()
 
     const err = await capturePromiseError(adapter.get('test-url'))
-    expect(err).to.be.instanceOf(Error)
-    expect(isTimeoutError(err)).to.be.true
-    expect(isHttpError(err)).to.be.true
-    expect(err.toString()).to.eql('Tequilapi error: timeout of 1ms exceeded (path="test-url")')
+    expect(err).to.be.instanceOf(TequilapiError)
+    expect(err.isTimeoutError()).to.be.true
```
mipo47: IDE shows warning here, err need to be casted to TequilapiError

src/libraries/mysterium-tequilapi/tequilapi-error.js
donce: This is used by `toString()`

donce: Added spec to clarify that.

PR #287, opened 2018-08-23
bugfix/MYS-178 favorite button select

shroomist:
disabling text selection in app


ignasbernotas:
@shroomist Missing cursor css

PR #294, opened 2018-08-27
Removal of email field in feedback form

ignasbernotas:


src/app/bug-reporting/feedback-form.js
```diff
@@ -18,6 +18,8 @@
 // @flow
 import RavenJs from 'raven-js'
 
+const FORM_EMAIL_VALUE = 'vpn.feedback@mysterium.network'
```
mipo47: Would be nice to move all constants to single file later,
it will be hard to find this line, 
most of our constants are in dependencies folders now

ignasbernotas: This constant will not be used for anything, this is just a placeholder value for the form field. The email field is required and needs to be filled with something.

shroomist: please move this as a param for feedbackForm constructor.
then pass it down from DI.
thanks.

ignasbernotas:
@mipo47 @shroomist @donce fixed

PR #297, opened 2018-08-28
MYS 196 temporary shown error

mipo47:
Double check mysterium client - call healthcheck at application start even if service is installed and started

### DIFF
--- a/src/libraries/function-looper.js
+++ b/src/libraries/function-looper.js
@@ -30,7 +30,6 @@ class FunctionLooper {
   _func: AsyncFunctionWithoutParams
   _threshold: number
   _running: boolean = false
-  _stopping: boolean = false
   _errorSubscriber: Subscriber<Error> = new Subscriber()
   _currentExecutor: ?ThresholdExecutor
   _currentPromise: ?Promise<void>
@@ -44,10 +43,11 @@ class FunctionLooper {
     if (this.isRunning()) {
       return
     }
+    this._running = true
 
     const loop = async () => {
       // eslint-disable-next-line no-unmodified-loop-condition
-      while (this._running && !this._stopping) {
+      while (this._running) {
         this._currentExecutor = new ThresholdExecutor(this._func, this._threshold)
         this._currentPromise = this._currentExecutor.execute()
         try {
@@ -57,18 +57,12 @@ class FunctionLooper {
         }
       }
     }
-
-    this._running = true
     loop()
   }
 
   async stop (): Promise<void> {
-    this._stopping = true
-
-    await this._waitForStartedPromise()
-
     this._running = false
-    this._stopping = false
+    await this._waitForStartedPromise()
   }
 
   isRunning (): boolean {
--- a/src/libraries/mysterium-client/service-manager/service-manager-process.js
+++ b/src/libraries/mysterium-client/service-manager/service-manager-process.js
@@ -57,10 +57,17 @@ class ServiceManagerProcess implements Process {
 
   async start (): Promise<void> {
     const state = await this._serviceManager.getServiceState()
-    if (state === SERVICE_STATE.RUNNING) {
-      return
+    let healthCheckWorks = false
+    try {
+      await this._tequilapi.healthCheck()
+      healthCheckWorks = true
+    } catch (e) {
+      logger.error('Unable to healthcheck while starting process', e)
     }
 
+    if (state === SERVICE_STATE.RUNNING && healthCheckWorks) {
+      return
+    }
     await this._repair(state)
   }
 
@@ -108,6 +115,10 @@ class ServiceManagerProcess implements Process {
   }
 
   async _waitForHealthCheck (): Promise<void> {
+    if (!this._monitoring.isStarted) {
+      this._monitoring.start()
+    }
+
     let resolveAndClearTimer: ?StatusCallback
     await new Promise((resolve, reject) => {
       const rejectTimer = setTimeout(() => reject(new Error('Unable to start service')), SERVICE_INIT_TIME)
--- a/src/libraries/mysterium-client/system.js
+++ b/src/libraries/mysterium-client/system.js
@@ -21,6 +21,7 @@ import { exec } from 'child_process'
 import sudo from 'sudo-prompt'
 import { promisify } from 'util'
 import fs from 'fs'
+import logger from '../../app/logger'
 
 const writeFile = promisify(fs.writeFile)
 
@@ -98,8 +99,9 @@ class OSSystem implements System {
           reject(ensureError(error))
           return
         }
-
-        resolve(stdout.trim())
+        const result = stdout.trim()
+        logger.info('user exec', command, result)
+        resolve(result)
       })
     })
   }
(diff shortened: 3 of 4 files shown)


src/libraries/function-looper.js
```diff
@@ -44,10 +43,11 @@ class FunctionLooper {
     if (this.isRunning()) {
       return
     }
+    this._running = true
```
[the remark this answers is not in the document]
ignasbernotas: Because if `this.isRunning()` returns true (e.g. it was stopped), the loop won't start again.

src/libraries/function-looper.js
```diff
@@ -44,10 +43,11 @@ class FunctionLooper {
...
 
     const loop = async () => {
       // eslint-disable-next-line no-unmodified-loop-condition
-      while (this._running && !this._stopping) {
+      while (this._running) {
```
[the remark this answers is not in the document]
mipo47: in stop method

src/libraries/function-looper.js
```diff
@@ -44,10 +43,11 @@ class FunctionLooper {
...
         this._currentExecutor = new ThresholdExecutor(this._func, this._threshold)
         this._currentPromise = this._currentExecutor.execute()
         try {
```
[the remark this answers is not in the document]
mipo47: we set it to false in stop method

src/libraries/function-looper.js
```diff
@@ -57,18 +57,12 @@ class FunctionLooper {
...
   }
 
   async stop (): Promise<void> {
-    this._stopping = true
-
-    await this._waitForStartedPromise()
-
     this._running = false
-    this._stopping = false
+    await this._waitForStartedPromise()
```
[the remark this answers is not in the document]
mipo47: that's the way to ensure loop is ended

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -108,6 +115,10 @@ class ServiceManagerProcess implements Process {
   }
 
   async _waitForHealthCheck (): Promise<void> {
+    if (!this._monitoring.isStarted) {
```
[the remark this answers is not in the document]
mipo47: fixed

src/app/mysterion.js
```diff
@@ -405,13 +405,13 @@ class Mysterion {
...
 
-  _startProcess () {
+  async _startProcess () {
     const cacheLogs = (level, data) => {
       this._mysteriumProcessLogCache.pushToLevel(level, data)
     }
 
     logInfo("Starting 'mysterium_client' process")
-    this._process.start()
+    await this._process.start()
```
ignasbernotas: ?? Is this legit? awaiting, but calling then/catch methods?

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -108,9 +115,14 @@ class ServiceManagerProcess implements Process {
...
+      await this._tequilapi.healthCheck()
+      return
```
ignasbernotas: unnecessary return?

src/app/mysterion.js
donce: Yes - you're basically awaiting both actions to suceed - process start plus logging.

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -57,10 +57,17 @@ class ServiceManagerProcess implements Process {
...
   async start (): Promise<void> {
     const state = await this._serviceManager.getServiceState()
-    if (state === SERVICE_STATE.RUNNING) {
-      return
+    let healthCheckWorks = false
+    try {
+      await this._tequilapi.healthCheck()
+      healthCheckWorks = true
+    } catch (e) {
+      logger.error('Unable to healthcheck while starting process', e)
```
donce: We can leave this method fewer responsibilieties (for SRP) by extracting healthcheckWorks implementation:
```js
...
const healtCheckWorks = this._healthCheck()
if (!healthCheckWorks) {
  logger.error('Unable to healthcheck while starting process', e)
}
...

async _healthCheck(): Boolean {
  try {
    await this._tequilapi.healthCheck()
    return true
  } catch (e) {
    return false
  }
}
```

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -57,10 +57,17 @@ class ServiceManagerProcess implements Process {
 
   async start (): Promise<void> {
     const state = await this._serviceManager.getServiceState()
-    if (state === SERVICE_STATE.RUNNING) {
-      return
+    let healthCheckWorks = false
```
donce: ?? This implementation is for Windows only - can we make this more abstract so that this would ensure that OS X will wait for healthcheck as well?

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -108,9 +115,14 @@ class ServiceManagerProcess implements Process {
   }
 
   async _waitForHealthCheck (): Promise<void> {
+    if (!this._monitoring.isStarted) {
+      await this._tequilapi.healthCheck()
```
donce: !! This can throw exception is healthcheck fails, and result will be thrown up to `Mysterion._startProcess()`, where it's not handled - this will result in unfinished bootstrap.

src/libraries/mysterium-client/monitoring.js
```diff
@@ -132,5 +141,26 @@ class Monitoring {
...
 
+function waitForStatusUp (tequilapi: TequilapiClient, timeout: number = healthCheckWaitTimeout): Promise<void> {
+  const monitoring = new Monitoring(tequilapi)
+  const statusUpAsync = promisify(monitoring.onStatusUp.bind(monitoring))
+  monitoring.start()
+
+  return Promise.race([
+    throwErrorAfterTimeout(timeout),
+    statusUpAsync()
+  ]).catch(e => {
```
shroomist: we can throw out this catch block

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -22,16 +22,21 @@ import type { LogCallback, Process } from '../index'
...
 import ServiceManager, { SERVICE_STATE } from './service-manager'
 import type { ServiceState } from './service-manager'
 
 /***
- * Time in milliseconds required to fully activate Mysterium client
+ * Time in milliseconds required to fully activate Mysterium client after restart
  * @type {number}
  */
-const SERVICE_INIT_TIME = 8000
+const SERVICE_INIT_TIME = 1500 * 4
```
shroomist: ?? what is 1500 in this math?
if its `SERVICE_CHECK_TIME` then we might as well use that const

test/unit/libraries/mysterium-client/monitoring.spec.js
```diff
@@ -144,4 +155,63 @@ describe('Monitoring', () => {
...
+      const waitPromise = waitForStatusUp(tequilapiClient, 15000)
+      await nextTick()
+      for (let i = 0; i < 10; i++) {
+        expect(tequilapiClient.healthCheckCallCount).to.be.eql(i + 1)
+        await tickWithDelay(1500)
+      }
+      await captureAsyncError(() => waitPromise)
+      expect(tequilapiClient.healthCheckCallCount).to.be.eql(11)
+    })
+    // it('stops calling healthcheck after first success', async () => {
```
shroomist: !! dead code?

PR #300, opened 2018-08-30
feature/MYS-69 send app_start & app_start_success

shroomist:
screen shot 2018-08-31 at 10 03 22

### DIFF
--- a/src/app/statistics/startup-event-tracker.js
+++ b/src/app/statistics/startup-event-tracker.js
@@ -21,7 +21,9 @@ import os from 'os'
 import type { EventSender } from './event-sender'
 import osName from 'os-name'
 
-const STARTUP_EVENT_NAME = 'startup'
+const STARTUP_EVENT_NAME = 'runtime_environment_details'
+const APP_START_EVENT_NAME = 'app_start'
+const APP_START_SUCCESS_EVENT_NAME = 'app_start_success'
 
 /**
  * Sends startup event.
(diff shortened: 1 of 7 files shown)


src/app/statistics/startup-event-tracker.js
```diff
@@ -22,6 +22,8 @@ import type { EventSender } from './event-sender'
 import osName from 'os-name'
 
 const STARTUP_EVENT_NAME = 'startup'
+const APP_START_EVENT_NAME = 'app_start'
```
donce: ?? `startup` and `app_start` are two very similar names - it's hard to tell the difference. Can we rename them somehow to highlight the difference?

Why do we have 2 different but similar events - because one has identity and another one doesn't? If so, maybe `startup` event can be renamed to `startup_identity`?

src/app/statistics/startup-event-tracker.js
```diff
@@ -22,6 +22,8 @@ import type { EventSender } from './event-sender'
...
+const APP_START_SUCCESS_EVENT_NAME = 'app_start_success'
```
donce: ?? Do you have graphs in ELK which calculate how many successes and/or failures happened?

src/app/statistics/startup-event-tracker.js
shroomist: that would break the graphs we have already that rely on `startup` event name.

src/app/statistics/startup-event-tracker.js
shroomist: added in description. tested only once :)

PR #306, opened 2018-09-03
Update some dev dependencies

donce:


donce:
Now I'm really done :D

PR #308, opened 2018-09-04
Update some dev dependencies

donce:


package.json
```diff
@@ -144,7 +144,7 @@
     "css-loader": "^0.28.11",
     "del": "^3.0.0",
     "devtron": "^1.4.0",
-    "electron": "^1.8.2",
+    "electron": "^1.8.8",
```
mipo47: ?? Latest stable version is v2.0.8, why we are not using it?

donce: Because it's not that easy to upgrade everything at once - builds start to fail. That what @ignasbernotas will be working on.

PR #311, opened 2018-09-05
Refactor/webpack v4

ignasbernotas:
Upgraded to webpack 4


package.json
```diff
@@ -186,8 +186,8 @@
     "vue-style-loader": "^4.1.2",
     "vue-svg-loader": "^0.8.0",
     "vue-template-compiler": "^2.5.17",
-    "webpack": "^3.11.0",
-    "webpack-dev-server": "^2.11.1",
+    "webpack": "^4.17.2",
```
donce: !! `karma-webpack` has to be updated as well - `yarn check` shows that
```
error "karma-webpack#webpack@^2.0.0 || ^3.0.0" doesn't satisfy found match of "webpack@4.17.2"
```

package.json
```diff
@@ -160,7 +160,7 @@
     "eslint-plugin-promise": "^4.0.0",
     "eslint-plugin-standard": "^3.1.0",
     "eslint-plugin-vue": "^4.7.1",
-    "extract-text-webpack-plugin": "^3.0.0",
+    "extract-text-webpack-plugin": "^4.0.0-beta.0",
```
donce: !! `mini-css-extract-plugin` should be used instead.

PR #315, opened 2018-09-06
MYS-155 increase test coverage

mipo47:


test/unit/app/terms/index.spec.js
```diff
@@ -0,0 +1,139 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// @flow
+
+import mock from 'mock-fs'
```
shroomist: I would prefer that we don't mock fs, but that's personal.
the test would have had more validity if we tested with actual files.

test/unit/app/terms/index.spec.js
```diff
@@ -0,0 +1,139 @@
...
+      terms.load()
+      terms.accept()
+
+      const writtenFileData = mock.getMockRoot()
+        ._items.TERMS_EXPORT_PATH
+        ._items['terms.html']
+        ._content
+
+      logger.info(writtenFileData)
+      expect(writtenFileData).to.have.lengthOf(TERMS_HTML.length)
```
shroomist: length is good, we can also check for content matching.

mipo47: fs-mock encodes content in strange way, decoding will be too hard for this simple test

test/unit/app/terms/index.spec.js
```diff
@@ -0,0 +1,139 @@
...
+    })
+
+    it('returns false when accepted contents does not matches terms', () => {
+      mock.restore()
+      mock({
+        [TERMS_EXPORT_PATH]: {
+          'terms.html': TERMS_HTML + '1'
+        }
+      })
+      terms.termsHtml = TERMS_HTML + '2'
```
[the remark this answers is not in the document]
mipo47: isAccepted will load mocked file and compare it with field termsHtml

test/unit/app/terms/index.spec.js
```diff
@@ -134,6 +135,10 @@ describe('Terms', () => {
       })
       expect(error).to.be.an('error')
       expect(content).to.be.undefined
+
+      if (error instanceof Error) {
```
shroomist: !! I think it's better to type cast `error = (error: Error)` and remove this condition.
otherwise, if error wasn't `instanceof Error` this test would pass, but should fail

src/main/dependencies/modules/mysterium-client.js
```diff
@@ -100,7 +100,7 @@ function bootstrap (container: Container) {
     (config: ClientConfig, platform: string, serviceManager: ServiceManager) => {
       switch (platform) {
         case OSX:
-          return new LaunchDaemonInstaller(config)
+          return new LaunchDaemonInstaller(config, new OSSystem())
```
shroomist: it is beautiful when installer signatures match.
```
LaunchDaemonInstaller(new OSSystem(), config) 
ServiceManagerInstaller(new OSSystem(), config, serviceManager)
```

test/unit/app/terms/index.spec.js
```diff
@@ -0,0 +1,139 @@
...
+  describe('.load', () => {
+    it('loads terms html from file', () => {
+      terms.load()
+      expect(terms.termsHtml).to.be.eql(TERMS_HTML)
+    })
+
+    it('throws exception when terms file is not found', () => {
+      mock.restore()
+      const error = captureError(() => terms.load())
+      expect(error).to.be.an('error')
```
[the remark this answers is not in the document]
shroomist: we're still not checking for specific error message of specific error type here.

PR #324, opened 2018-09-11
[MYS-223] Unify metric syncing

donce:
- Before this, both sync and async messages were sent for metrics synchronisation.
- Renderer fetches metrics synchronously using sync communication (didn't change this), so asynchronous communications are not needed anymore.

download


src/app/bug-reporting/metrics/bug-reporter-metrics-store.js
```diff
@@ -0,0 +1,59 @@
...
+    return this._metrics.get(metric)
+  }
+
+  setWithCurrentDateTime (metric: Metric) {
+    this.set(metric, getCurrentTimeISOFormat())
+  }
+
+  getMetrics (): RavenData {
+    const data = { tags: {}, extra: {} }
+    data.tags = this._getValues((Object.values(TAGS): any))
```
shroomist: ?? I'm curious why do we need casting to `any` here? 
shouldn't it be casted to `Array<Metric>`?

src/app/communication/dto.js
```diff
@@ -56,3 +57,8 @@ export type LogDTO = {
   level: LogLevel,
   data: mixed
 }
+
+export type MetricValueDto = {
```
shroomist: I think we can drop `Value` and name this `MetricDTO`

src/app/communication/renderer-communication.js
```diff
@@ -85,11 +84,11 @@ class RendererCommunication implements MapSyncCommunication<Metric> {
     return this._send(messages.TOGGLE_FAVORITE_PROVIDER, data)
   }
 
-  sendMapUpdate (data: MapSyncDTO<Metric>): void {
+  sendMapUpdate (data: MetricValueDto): void {
```
shroomist: Can we also make this method name more descriptive?
`sendMetricUpdate` reads better to me

test/helpers/communication/fake-sync-renderer-communication.js
```diff
@@ -0,0 +1,54 @@
...
+    mysterium_process: { info: 'mysterium info', error: 'mysterium error' },
+    frontend: { info: 'frontend info', error: 'frontend error' }
+  }
+  mockedMetrics: RavenData = {
+    tags: { [TAGS.CLIENT_RUNNING]: true },
+    extra: {}
+  }
+  sentMetric: ?MetricValueDto = null
+
+  getSerializedCaches () {
```
shroomist: !! how about return type here?

test/unit/app/bug-reporting/metrics/bug-reporter-metrics-store.spec.js
```diff
@@ -0,0 +1,90 @@
...
+      const tagKeys = Object.keys(data.tags)
+      const extraKeys = Object.keys(data.extra)
+
+      expect(tagKeys.length).to.eql(Object.values(TAGS).length)
+      expect(extraKeys.length).to.eql(Object.values(EXTRA).length)
+
+      for (let tagKey of Object.values(TAGS)) {
+        expect(tagKeys).to.contain(tagKey)
+        // $FlowFixMe
+        expect(data.tags[tagKey]).to.deep.equal(store.get(tagKey) || NOT_SET)
```
shroomist: ?? 😮 why ` || NOT_SET` here?

src/app/bug-reporting/metrics/bug-reporter-metrics-store.js
```diff
@@ -0,0 +1,59 @@
...
+ */
+class BugReporterMetricsStore implements BugReporterMetrics {
+  _metrics: Map<Metric, mixed> = new Map()
+
+  set (metric: Metric, value: mixed) {
+    this._metrics.set(metric, value)
+  }
+
+  get (metric: Metric): mixed {
+    return this._metrics.get(metric)
```
shroomist: maybe we should do `|| NOT_SET` here?

src/app/bug-reporting/metrics/bug-reporter-metrics-store.js
donce: We can't do that, because `Object.values` returns `Array<mixed>`. Even though it is `Array<Metric>`, Flow is not aware of that.

src/app/communication/renderer-communication.js
donce: These messages aren't actually used anymore, will remove these.

test/unit/app/bug-reporting/metrics/bug-reporter-metrics-store.spec.js
donce: @mipo47 it seems to be your code, could you explain why? :D

src/app/bug-reporting/metrics/bug-reporter-metrics-store.js
shroomist: lets go all the way ;) `data.tags = this._getValues(((Object.values(TAGS): any): Array<Metric>))`

src/app/bug-reporting/metrics/bug-reporter-metrics-store.js
```diff
@@ -0,0 +1,59 @@
...
+ */
+class BugReporterMetricsStore implements BugReporterMetrics {
+  _metrics: Map<Metric, mixed> = new Map()
+
+  set (metric: Metric, value: mixed) {
+    this._metrics.set(metric, value)
+  }
+
+  get (metric: Metric): mixed {
+    return this._metrics.get(metric) || 'N/A'
```
shroomist: ?? shy not `|| NOT_SET`?

PR #326, opened 2018-09-12
MYS 192 runtime type checking

mipo47:
Some tests still contains runtime type errors, that's why checking is turned off now.

To turn type checking open .babelrc and change: 
"assert": false to "assert": true

Check out this branch and run 'yarn' to ensure, that your node-js version is supported


.babelrc
```diff
@@ -12,7 +12,14 @@
         }],
         "stage-0"
       ],
-      "plugins": ["istanbul"]
+      "plugins": [
+        "istanbul",
+        "transform-decorators-legacy",
```
donce: ?? Is this needed by flow-runtime?

src/app/statistics/elk-collector.js
```diff
@@ -20,7 +20,7 @@ import axios from 'axios'
 import type { Event, EventCollector } from './events'
 
 class ElkCollector implements EventCollector {
-  _axiosApi: axios.Axios
+  _axiosApi: any
```
donce: Maybe we can create a local type definition for methods we use?

src/libraries/mysterium-tequilapi/dto/identity-registration.js
```diff
@@ -16,6 +16,7 @@
  */
 
 // @flow
+// @flow-runtime
```
donce: ?? Do these work with flow-runtime disabled?

src/renderer/store/modules/connection.js
```diff
@@ -37,11 +37,11 @@ import type { Provider } from '../../../app/connection/provider'
 import messages from '../../../app/messages'
 
 type ConnectionStore = {
-  ip: ?string,
-  location: ?ConsumerLocationDTO,
+  ip?: ?string,
```
donce: ?? Why 2 question marks? One is usually enough.

mipo47: one question mark to allow skip property,
another one to make property nullable
(no way to make it with one question mark)

src/libraries/mysterium-tequilapi/dto/identity-registration.js
mipo47: hopefully no, I'll remove it

src/app/statistics/elk-collector.js
mipo47: sounds like great future task

.babelrc
mipo47: Can't remember why I've added it, but now it's not necessary

src/renderer/store/modules/connection.js
donce: Maybe we can have `ip: ?string` and initialise this with `{ ip: null }` ?

src/app/communication/dto.js
```diff
@@ -22,7 +22,7 @@ import type { Country } from '../countries'
 import type { Metric } from '../bug-reporting/metrics/metrics'
 
 export type ConnectionStatusChangeDTO = {
-  oldStatus: ConnectionStatus,
+  oldStatus: ?ConnectionStatus,
   newStatus: ConnectionStatus
```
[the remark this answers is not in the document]
mipo47: before first request you have no idea about status, that's why it's nullable.
New status is always latest status, can't be null

src/renderer/store/modules/connection.js
mipo47: don't want to change current code, so I just fixed type definition

PR #329, opened 2018-09-17
[MYS-244] Do not send common http errors

donce:
Proposal fetching happens every 5s when internet is down.
Since day has 24 * 60 * 60 = 86400 seconds.
**one day** of application running without internet triggers **17280 events**.
Results:

screen shot 2018-09-17 at 11 06 38
screen shot 2018-09-17 at 11 06 42

We are above our usage limit, therefore we cannot log all requests errors :(


src/app/mysterium-tequilapi/adapters/logger-adapter.js
```diff
@@ -18,19 +18,18 @@
 // @flow
 
 import type { HttpInterface, HttpQueryParams } from '../../../libraries/mysterium-tequilapi/adapters/interface'
-import type { BugReporter } from '../../bug-reporting/interface'
-import TequilapiError from '../../../libraries/mysterium-tequilapi/tequilapi-error'
+import { Logger } from '../../logger'
 
 /**
  * Delegates to other 'HttpInterface' and captures errors to 'BugReporter'.
```
mipo47: !! BugReporter is thrown away, change description

src/app/mysterium-tequilapi/adapters/logger-adapter.js
```diff
@@ -57,12 +56,10 @@ class BugReporterAdapter implements HttpInterface {
     try {
       return await func()
     } catch (err) {
-      if (err instanceof TequilapiError) {
-        this._bugReporter.captureInfoException(err)
-      }
+      this._logger.info(err)
```
mipo47: ?? maybe logger.error(), because we don't expect http errors

src/app/mysterium-vpn.js
```diff
@@ -445,11 +445,9 @@ class MysteriumVpn {
       logException('Proposal fetching failed', error)
     }
     this._communication.onProposalUpdateRequest(() => {
-      try {
-        this._proposalFetcher.fetch()
-      } catch (error) {
-        handleProposalFetchError(error)
-      }
+      this._proposalFetcher.fetch().catch((err: Error) => {
```
mipo47: async / await looks better

test/unit/app/mysterium-tequilapi/adapters/logger-adapter.spec.js
```diff
@@ -0,0 +1,161 @@
...
+import { beforeEach, describe, expect, it } from '../../../../helpers/dependencies'
+import LoggerAdapter from '../../../../../src/app/mysterium-tequilapi/adapters/logger-adapter'
+import type { HttpInterface } from '../../../../../src/libraries/mysterium-tequilapi/adapters/interface'
+import MockHttpAdapter from '../../../../helpers/adapters/MockHttpAdapter'
+import type { StringLogger } from '../../../../../src/app/logging/string-logger'
+import { Logger } from '../../../../../src/app/logger'
+import { captureAsyncError } from '../../../../helpers/utils'
+
+class MockStringLogger implements StringLogger {
+  logs: {level: string, message: string}[] = []
```
mipo47: I like it (compact type definition)

test/unit/app/mysterium-tequilapi/adapters/logger-adapter.spec.js
```diff
@@ -0,0 +1,161 @@
...
+      expect(response).to.eql(mockResponse)
+
+      expect(mockAdapter.lastPath).to.eql('path')
+      expect(mockAdapter.lastQuery).to.eql({ key: 'value' })
+      expect(mockAdapter.lastTimeout).to.eql(5)
+    })
+
+    it('logs error', async () => {
+      mockAdapter.mockError = new Error('mock error')
+      await captureAsyncError(() => adapter.get('path'))
```
mipo47: Assert returned error

src/app/mysterium-vpn.js
shroomist: we're only catching here, so this looks shorter

test/unit/app/mysterium-tequilapi/adapters/logger-adapter.spec.js
shroomist: ?

donce:
> we're only catching here, so this looks shorter

it's not just because it's shorter - `.fetch` is async, so try/catch didn't actually work, because it's not `async` function :D

> ?

?

src/app/mysterium-tequilapi/adapters/logger-adapter.js
donce: We do expect errors - i.e. if user has no internet, all proposal fetches will fail, and this is expected.

src/app/mysterium-vpn.js
donce: Also, we can't await here - communication method expect sync callback.

src/app/mysterium-tequilapi/adapters/logger-adapter.js
donce: Fixed.

test/unit/app/mysterium-tequilapi/adapters/logger-adapter.spec.js
mipo47: await captureAsyncError() returns error, you should check error message

PR #330, opened 2018-09-18
Feature/mvpn 217 tequilapi pkg

shroomist:
features https://www.npmjs.com/package/mysterium-tequilapi


package.json
```diff
@@ -96,6 +96,7 @@
     }
   },
   "dependencies": {
+    "@babel/runtime": "^7.0.0",
```
donce: Why do we need this?

PR #331, opened 2018-09-18
hotfix/MVPN-240 registrationLink

shroomist:


src/dependencies/payments.js
```diff
@@ -17,13 +17,20 @@
...
     process.env.NODE_ENV === 'development'
       ? 'http://wallet-dev.mysterium.network/'
-      : 'http://walletx.mysterium.network/'
+      : 'http://wallet.mysterium.network/'
+  )
+
+  container.service(
+    'getPaymentLink',
+    ['paymentBaseUrl'],
+    (paymentBaseUrl) => getPaymentLink.bind(null, paymentBaseUrl)
```
donce: ?? Function as dependency seems strange for me, especially when using `bind` - it seemed less confusing before.
It's unclear what params returned function actually requires. To see what params `getPaymentLink` requires, you have to jump twice in code :/

If we want to have this in DI, maybe lambda function with explicit param here seems more clear?
```js
paymentBaseUrl => {
  return registration => { getPaymentLink(paymentBaseUrl, registration) }
}
```

Or a small class? I know you hate classes, but at least they show you required params :D

```js
class PaymentLinkGenerator {
  constructor (baseUrl: string) {...}
  generate (registration: string): string {...}
}
```

PR #332, opened 2018-09-18
Refactor user settings store

donce:
- Add `Observer` class
- Single `UserSettingsStore` interface in both processes with different implementations: `UserSettingsStorage` and `UserSettingsProxy`

Motivation:

* improve listening for changes (observing):
    * allow observing single property instead of whole settings where it’s not needed
        * features need to know less
        * fewer irrelevant updates arrive
    * notify observers only when setting actually changes
        * fewer irrelevant updates arrive
* uncouple features from communication:
    * same domain logic works in either process
        * ability to move/reuse features around processes
    * avoid sending `USER_SETTINGS_REQUEST` message in each observer
    * avoid adding IPC listeners for each value observer - send only once
* Ability later on to easily move real storage to renderer process if needed
* Ability later on to abstract this into something more general, i.e. `SyncedObservable`


test/unit/app/user-settings/storage.spec.js
```diff
@@ -0,0 +1,103 @@
...
+
+    it('exports a valid json file', async () => {
+      const settings: UserSettings = {
+        showDisconnectNotifications: false,
+        favoriteProviders: new Set(['id_123'])
+      }
+      await saveSettings(saveSettingsPath, settings)
+      const data = readFileSync(saveSettingsPath, { encoding: 'utf8' })
+
+      expect(data.toString()).to.eql(
```
ignasbernotas: Could use string interpolation here?

test/unit/libraries/observable.spec.js
```diff
@@ -0,0 +1,53 @@
...
+
+// @flow
+
+import { beforeEach, describe, expect, it } from '../../helpers/dependencies'
+import { Observable } from '../../../src/libraries/observable'
+import { RepeatableCallbackRecorder } from '../../helpers/utils'
+
+describe('Observable', () => {
+  describe('.subscribe', () => {
+    let observable
```
mipo47: define type, it will help IDE to autocomplete

src/app/user-settings/observable-user-settings.js
```diff
@@ -0,0 +1,92 @@
...
+    const settings = getDefaultSettings()
+    const observables = {}
+    this._getProperties().forEach(setting => {
+      observables[setting] = new Observable(settings[setting])
+    })
+    return (observables: any)
+  }
+
+  _getProperties (): UserSettingName[] {
+    return (Object.values(userSettingName): any)
```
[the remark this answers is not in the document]
donce: Mixed cannot be casted from `mixed` to `UserSettingName[]` directly - that why we need `any`.

PR #335, opened 2018-09-20
Updated Electron to v3, spectron to v4

ignasbernotas:


package.json
```diff
@@ -183,7 +183,7 @@
     "node-sass": "^4.9.3",
     "require-dir": "^1.0.0",
     "sass-loader": "^7.1.0",
-    "spectron": "^3.8.0",
+    "spectron": "^4.0.0",
```
donce: Does this fix spectron as well?

ignasbernotas: Don't think so. This is just a version bump.

PR #341, opened 2018-09-25
IP hidden by default

ignasbernotas:
kapture 2018-09-25 at 22 38 53

### DIFF
--- a/src/renderer/pages/vpn.vue
+++ b/src/renderer/pages/vpn.vue
@@ -25,7 +25,7 @@
         <div
           class="control__location">
           current IP:
-          <span :class="{'text-blurry': ip}">{{ ip ? ip : 'Refreshing..' }}</span>
+          <span :class="{'text-blurry': ip}">{{ ip || 'Refreshing..' }}</span>
         </div>
       </div>
 
--- a/src/renderer/pages/vpn.vue
+++ b/src/renderer/pages/vpn.vue
@@ -23,8 +23,10 @@
           :class="{'is-grey':statusCode===-1}"
           v-text="statusTitle"/>
         <div
-          class="control__location"
-          v-if="ip">current IP: {{ ip }}</div>
+          class="control__location">
+          current IP:
+          <span :class="{'text-blurry': ip}">{{ ip ? ip : 'Refreshing..' }}</span>
+        </div>
       </div>
 
       <div class="control__bottom">
--- a/src/renderer/store/modules/connection.js
+++ b/src/renderer/store/modules/connection.js
@@ -68,8 +68,6 @@ class ActionLooperConfig {
 const defaultStatistics = {
 }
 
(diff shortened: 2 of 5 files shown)


src/renderer/pages/vpn.vue
```diff
@@ -23,8 +23,10 @@
           :class="{'is-grey':statusCode===-1}"
           v-text="statusTitle"/>
         <div
-          class="control__location"
-          v-if="ip">current IP: {{ ip }}</div>
+          class="control__location">
+          current IP:
+          <span :class="{'text-blurry': ip}">{{ ip ? ip : 'Refreshing..' }}</span>
```
donce: `ip || 'Refreshing...'` ?

src/renderer/store/modules/connection.js
```diff
@@ -68,7 +68,7 @@ class ActionLooperConfig {
 const defaultStatistics = {
 }
 
-const IP_REFRESHING = 'Refreshing...'
+const IP_UNAVAILABLE = null
```
donce: Maybe we can skip this constant?
If you want to keep this constant, we should reuse it in `const state = { ip: null, ... }`

src/renderer/store/modules/connection.js
```diff
@@ -191,13 +191,13 @@ function actionsFactory (
...
       if (newStatus === ConnectionStatusEnum.CONNECTED) {
-        commit(type.CONNECTION_IP, IP_REFRESHING)
+        commit(type.CONNECTION_IP, IP_UNAVAILABLE)
         const statisticsLooperConfig =
           new ActionLooperConfig(type.CONNECTION_STATISTICS, config.statisticsUpdateThreshold)
         await dispatch(type.START_ACTION_LOOPING, statisticsLooperConfig)
       }
       if (newStatus === ConnectionStatusEnum.NOT_CONNECTED) {
-        commit(type.CONNECTION_IP, IP_REFRESHING)
+        commit(type.CONNECTION_IP, IP_UNAVAILABLE)
```
donce: !! This broke the tests - please update them.

src/renderer/pages/vpn.vue
```diff
@@ -23,8 +23,10 @@
           :class="{'is-grey':statusCode===-1}"
           v-text="statusTitle"/>
         <div
-          class="control__location"
-          v-if="ip">current IP: {{ ip }}</div>
+          class="control__location">
+          current IP:
+          <span :class="{'text-blurry': ip}">{{ ip || 'Refreshing..' }}</span>
```
donce: Previously, we didn't show this label at all until first ip fetch passed - now we'll show 'current IP: Refreshing..'. This is probably intentional?

ignasbernotas: Yep:)

PR #345, opened 2018-09-27
Yarn CI command lists outdated packages

ignasbernotas:
screen shot 2018-09-27 at 16 07 38


util_scripts/outdated-dependencies.js
```diff
@@ -29,6 +29,18 @@ const printPackageInfo = ({ name, current, latest }) => {
   console.log('\x1b[31m%s\x1b[0m - current: %s, latest: %s', name, current, latest)
 }
 
+const listOutdatedPackages = (packages) => {
+  packages.data.body.forEach((pkg) => {
+    if (needsToBeUpdated(pkg[1], pkg[3])) {
```
donce: pkg[1], pkg[3], pkg[0] isn't quite readable, I would extract them to named variables.

util_scripts/outdated-dependencies.js
```diff
@@ -0,0 +1,60 @@
...
+
+  return false
+}
+
+const printPackageInfo = ({ name, current, latest }) => {
+  console.log('\x1b[31m%s\x1b[0m - current: %s, latest: %s', name, current, latest)
+}
+
+const listOutdatedPackages = (packages) => {
+  packages.data.body.forEach((pkg) => {
```
shroomist: how about doing it more pure?
```js
const getOutdatedPackagesString = (packages) => {
return packages.data.body
    .filter((pkg) => needsToBeUpdated(pkg[1], pkg[3]))
    .map((pkg) => beautifySerialize{
         name: pkg[0],
         current: pkg[1],
         latest: pkg[3]
       })
    .join('\n')

}
...
const beautifySerialize = ({ name, current, latest }) => {
   return `\x1b[31m%${name}x1b[0m - current: ${current}, latest: ${latest}`
 }
const packages = JSON.parse(json)
const outdated = getOutdatedPackages(packages)
console.log(outdated)

ignasbernotas: This is an equivalent to a bash script, don't see much point in doing this

shroomist: ... but I've tried so hard coding it in github.. 😿

ignasbernotas:
@shroomist @donce please review again:)

PR #347, opened 2018-10-02
[MVPN-285] Move identity button to VPN page

donce:
Fixes these glitches:
kapture 2018-09-28 at 16 01 11
kapture 2018-09-28 at 15 59 30
screenshot 2018-09-28 at 15 54 14


src/renderer/components/identity-button.vue
```diff
@@ -0,0 +1,47 @@
...
+  -
+  - You should have received a copy of the GNU General Public License
+  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<template>
+  <div
+    class="identity-button"
+    :class="{'identity-button--registered': registered, 'identity-button--unregistered': !registered}"
+    @click="this.click">
```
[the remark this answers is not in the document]
donce: Nope, removed.

src/renderer/pages/vpn.vue
```diff
@@ -117,6 +125,21 @@ export default {
     },
     providerCountry () {
       return this.country ? this.country.code : null
+    },
+    registrationFetched () {
+      return this.registration != null
+    },
+    registration () {
+      return this.$store.state.identity.registration
```
shroomist: it's better to use `store.getters` in such cases

src/renderer/pages/vpn.vue
```diff
@@ -139,6 +162,20 @@ export default {
       if (countries.length < 1) this.bugReporter.captureInfoMessage('Renderer received empty countries list')
 
       this.countryList = countries
+    },
+    openPaymentsOrShowInstructions () {
+      if (this.registered) {
+        this.openPaymentsUrl()
+      } else {
```
shroomist: I think we favor the following 
```
      if (this.registered) {
         this.openPaymentsUrl()
         return
      }
      this.showInstructions()
```

src/renderer/store/modules/identity.js
```diff
@@ -63,17 +69,15 @@ const getters = {
   }
 }
 
-function factory (tequilapi: TequilapiClient, dependenciesContainer: Container) {
+function factory (bugReporter: BugReporter, communication: RendererCommunication) {
   return {
-    state,
+    state: stateFactory(),
```
shroomist: doesn't seem like we really need this `stateFactory()`

src/renderer/store/modules/main.js
```diff
@@ -36,13 +37,28 @@ const state: State = {
...
   navOpen: false,
+  identityMenuOpen: false,
   clientVersion: null,
   navVisible: true,
   errorMessage: null,
   error: null,
   showError: false
 }
 
+function stateFactory (): State {
```
shroomist: why?

donce: Because I've encountered multiple times that same state is persisted between different tests.

I was running multiple tests at once, and one of them changed state property, on which the other test depended, resulting in failing test - I've spend about an hour figuring out what was wrong.

src/renderer/store/modules/identity.js
donce: https://github.com/mysteriumnetwork/mysterium-vpn/pull/347#discussion_r221943039

src/renderer/pages/vpn.vue
donce: I do like this syntax more as well, especially where one default flow exists, i.e.
```js
createIfPossible () {
  if (!databaseExists()) {
    return
  }
  databaseCreate()
}

But since this method consists of two equal actions `openPaymentsOrShowInstructions`, `else` seems alright to me.

PR #348, opened 2018-10-02
Refactor vuex stores

donce:
- Converted `state` constants to factories. When using vuex in tests, the same `state` const is used in different test-cases, resulting in brain-fuck errors and a lot of time wasted. Factories isolate state for tests
- Encapsulated stores, exposing only what's necessary. Exposing `factory` instead of `getters`+`actions`+`mutations`+`state`.


src/renderer/app.vue
```diff
@@ -113,8 +113,6 @@ export default {
 
     // if the client was down, but now up, we need to unlock the identity once again
     this.rendererCommunication.healthcheckUp.on(() => {
-      this.$store.dispatch('setClientRunningState', true)
```
ignasbernotas: ?? Was this not used anywhere?

donce: Yep, it was just being set, but nobody used it.

PR #349, opened 2018-10-03
[MVPN-287] Extract CountryFlag component

donce:
Uncoupled country flag from dropdown.
This will be re-used for connection history.


src/app/countries/country-image-resolver.js
```diff
@@ -0,0 +1,60 @@
...
+
+  _getDefaultIconPath (): string {
+    return this._getIconPath('world')
+  }
+
+  _getIconPath (code: string): string {
+    return path.join('static', 'flags', code.toLowerCase() + '.svg')
+  }
+
+  _reportCountryIfUnique (code: string) {
```
shroomist: `_reportUnresolvedCountryByCode()`
country FLAG resolver shouldn't care about `isCountryKnown` and `_unresolvedCountryList`.  
resolving flag image should be a simple task without side effects.

I'd rather put encapsulate country (name and code) resolution in a class and do error handling there.
just imho anyway.

src/renderer/components/country-select.vue
```diff
@@ -39,38 +39,28 @@
...
-      <img
-        :src="imagePath(country.code)"
-        v-if="country"
-        class="countries__flag-svg">
-      <icon-world
-        class="countries__flag-svg"
-        v-if="!country"/>
-    </i>
+    <country-flag
+      :code="country ? country.code : null"
```
shroomist: can't we enforce that we get `country` object here?

ignasbernotas: Component doesn't need the full object:)

src/app/countries/country-image-resolver.js
donce: Renamed method to `_reportUnknownCountryCode`.

If we put this logic somewhere in `CountryFlag` component, then each component would report unknown country separately thus flooding the Sentry - having `unknownCountries` list in DI, we have single list instead of n lists.

src/renderer/components/country-select.vue
donce: Exactly @ignasbernotas point - the more unneeded info we pass to component, the more we couple that component to the outside world.

I.e. for connection-history, I'm not using `Country` DTO, so I would have to create a fake empty `Country` object.

donce:
Ping @shroomist

PR #350, opened 2018-10-04
WIP: [MVPN-153] OSX reinstall service when client version changes

ignasbernotas:
This PR kills the client when packaged version doesn't match the currently running client version.
It is missing tests, but I'll try to write as many as I can before I leave. In case I won't finish it, I'd like for you guys to finish it for us, since there are improvements that clean up the code base.


src/libraries/mysterium-client/index.js
```diff
@@ -36,6 +36,8 @@ interface Process {
 
   stop (): Promise<void>,
 
+  kill (): Promise<void>,
```
shroomist: ?? I don't see why we need both stop and kill

src/libraries/mysterium-client/version-check.js
```diff
@@ -0,0 +1,38 @@
...
+
+  constructor (tequilapi: TequilapiClient, packageVersion: ?string) {
+    this._tequilapi = tequilapi
+    this._packageVersion = packageVersion
+  }
+
+  async runningVersionMatchesPackageVersion (): Promise<boolean> {
+    const healthCheckResponse = await this._tequilapi.healthCheck()
+
+    return healthCheckResponse.version === this._packageVersion
```
shroomist: perhaps we could run a semver check here, that would also include the caret ranges.
see `npm semver`

.electron-vue/webpack.main.config.js
```diff
@@ -14,6 +14,10 @@ const features = require('./features')
 
 const BabiliWebpackPlugin = require('babili-webpack-plugin')
 
+if(!dependencies['mysterium-client-bin'].match(/^(\d+\.\d+\.\d+)$/)) {
+  throw new Error('mysterium-client-bin package must use an exact version.')
```
donce: ?? Shouldn't this be skipped if `clientVersionCheck` is disabled?

src/app/mysterium-vpn.js
```diff
@@ -165,10 +154,10 @@ class MysteriumVpn {
     this._communicationBindings.setCurrentIdentityForEventTracker(this._startupEventTracker)
     this._communicationBindings.syncCurrentIdentityForBugReporter(this._bugReporter)
 
-    this._communicationBindings.startRegistrationFetcherOnCurrentIdentity(
-      this._featureToggle,
-      this._registrationFetcher
-    )
+    if (featureToggle().paymentsAreEnabled()) {
+      this._communicationBindings.startRegistrationFetcherOnCurrentIdentity(this._registrationFetcher)
```
donce: 👍👍👍
Totally support moving feature check outside of feature! That's the whole idea - implement feature as it is enabled, but turn it on/off from outside!

src/app/mysterium-client/process-manager.js
```diff
@@ -0,0 +1,225 @@
...
+    this._monitoring = monitoring
+    this._communication = communication
+    this._logCache = logCache
+    this._versionCheck = versionCheck
+  }
+
+  async ensureInstallation () {
+    const needsInstallation = await this._installer.needsInstallation()
+    if (!needsInstallation) {
+      return true
```
donce: ?? Why do we return `true` here, isn't fact that function executed without error our success indicator?

src/app/mysterium-client/process-manager.js
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    try {
+      await this._installProcess()
+    } catch (error) {
+      const installed = await this._installer.needsInstallation()
+      if (!installed) {
+        bugReporter().captureErrorMessage('Process installation failed', error.message)
```
donce: ?? Can we capture exception here to have an actual stack-trace of error?

src/app/mysterium-client/process-manager.js
```diff
@@ -0,0 +1,225 @@
...
+    const needsInstallation = await this._installer.needsInstallation()
+    if (!needsInstallation) {
+      return true
+    }
+
+    try {
+      await this._installProcess()
+    } catch (error) {
+      const installed = await this._installer.needsInstallation()
+      if (!installed) {
```
donce: ?? Why do we need to check `needsInstallation` again - is it a happy path that `installProcess` throws an exception?

src/app/mysterium-client/process-manager.js
```diff
@@ -0,0 +1,225 @@
...
+    this._monitoring.onStatusDown(callback)
+  }
+
+  _startLogging () {
+    try {
+      this._process.setupLogging()
+      this._process.onLog(processLogLevels.INFO, (data) => this._logCache.pushToLevel(processLogLevels.INFO, data))
+      this._process.onLog(processLogLevels.ERROR, (data) => this._logCache.pushToLevel(processLogLevels.ERROR, data))
+    } catch (error) {
+      this._logError(`Failing to process logs. `, error.message)
```
donce: ?? "Failing to process logs" is very confusing - "processing logs" sounds like the actual log callback failed, but this `catch` does not catch logs processing, it only catches set-up of logging.
"Process logs setup failed" ?

src/app/mysterium-client/process-manager.js
```diff
@@ -0,0 +1,225 @@
...
+    this._monitoring.onStatusUp(callback)
+  }
+
+  onStatusDown (callback: DownCallback) {
+    this._monitoring.onStatusDown(callback)
+  }
+
+  _startLogging () {
+    try {
+      this._process.setupLogging()
```
donce: !! Since this is in `try catch`, I assume that this should have `await`, because otherwise there's not much this `catch` can actually catch.

src/main/helpers/feature-toggle.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+// @flow
+
+import dependencies from './../dependencies'
+import FeatureToggle from '../../app/features/feature-toggle'
+
+const featureToggle = (): FeatureToggle => {
+  return dependencies.get('featureToggle')
+}
```
donce: This is quite convenient to use in features, but we'll be forced to build DI containers in our tests for every class which uses FeatureToggle :/

src/libraries/mysterium-client/index.js
ignasbernotas: I understand the confusion although this is necessary as stop doesn't necessarily kill the process (in windows).

src/libraries/mysterium-client/version-check.js
ignasbernotas: We don't need it. We should always run the version of the client that the app is shipped with.

.electron-vue/webpack.main.config.js
```diff
@@ -14,6 +14,10 @@ const features = require('./features')
 
 const BabiliWebpackPlugin = require('babili-webpack-plugin')
 
+if(!dependencies['mysterium-client-bin'].match(/^(\d+\.\d+\.\d+)$/)) {
```
VaidotasSm: Minor - I think with regexp's since thay're very error prone it's good practice to have centralized util functions e.g. `regexps.isSemVer(...)` also same regexp is  used in `outdated-dependencies.js`

PR #351, opened 2018-10-05
[MVPN-203] WIP: Connection history

donce:
The frontend part is done (except for icon), so want to get this reviewed to be able to merge this as soon as blockers are gone.

screenshot 2018-10-05 at 09 37 48

TODO:
- [ ] Add custom menu icon (waiting on Andrius)
- [ ] Use real mysterium-client endpoint https://github.com/mysteriumnetwork/node/issues/433.

### DIFF
--- a/src/app/bug-reporting/tequilapi-client-with-metrics.js
+++ b/src/app/bug-reporting/tequilapi-client-with-metrics.js
@@ -116,6 +116,29 @@ class TequilapiClientWithMetrics implements TequilapiClient {
   async location (timeout: ?number): Promise<ConsumerLocationDTO> {
     return this._client.location(timeout)
   }
+
+  async connectionHistoryList () {
+    return [
+      {
+        id: '1',
+        identity: '0x3b03a513fba4bd4868edd340f77da0c920150f3e',
+        start: '2018.09.24 14:23:23',
+        status: 'Successful',
+        duration: '00:35:00',
+        sent: '1MB',
+        received: '5MB'
+      },
+      {
+        id: '2',
+        identity: '0x1b03b513fba4bd4868edd340f77da0c920150f0a',
+        start: '2018.09.24 14:23:23',
+        status: 'Successful',
+        duration: '00:35:00',
+        sent: '1MB',
+        received: '5MB'
+      }
+    ]
+  }
 }
 
 export default TequilapiClientWithMetrics
--- a/src/app/bug-reporting/tequilapi-client-with-metrics.js
+++ b/src/app/bug-reporting/tequilapi-client-with-metrics.js
@@ -33,6 +33,7 @@ import type { BugReporterMetrics } from './metrics/bug-reporter-metrics'
 import { METRICS } from './metrics/metrics'
 
 type ConnectionRecordDto = {
+  id: string,
   provider: {
     identity: string,
     country: string
@@ -132,6 +133,7 @@ class TequilapiClientWithMetrics implements TequilapiClient {
   async connectionHistoryList (): Promise<ConnectionRecordDto[]> {
     return [
       {
+        id: '30f610a0-c096-11e8-b371-ebde26989839',
         provider: {
           identity: '0x3b03a513fba4bd4868edd340f77da0c920150f3e',
           country: 'lt'
@@ -143,6 +145,7 @@ class TequilapiClientWithMetrics implements TequilapiClient {
         bytesReceived: 6000
       },
       {
+        id: '76fca3dc-28d0-4f00-b06e-a7d6050699ae',
         provider: {
           identity: '0x1b03b513fba4bd4868edd340f77da0c920150f0a',
           country: 'us'
@@ -154,6 +157,7 @@ class TequilapiClientWithMetrics implements TequilapiClient {
         bytesReceived: 6000
       },
       {
+        id: 'ffbfb796-5483-4a1b-82c8-10d6b85d4d62',
         provider: {
           identity: '0x1b03b513fba4bd4868edd340f77da0c920150f0a',
           country: 'us'
@@ -165,6 +169,7 @@ class TequilapiClientWithMetrics implements TequilapiClient {
         bytesReceived: 6000
       },
       {
+        id: '27f49b3a-070b-4d2e-881b-5163b9ab0a74',
         provider: {
           identity: '0x1b03b513fba4bd4868edd340f77da0c920150f0a',
           country: 'us'
@@ -176,6 +181,7 @@ class TequilapiClientWithMetrics implements TequilapiClient {
         bytesReceived: 6000
       },
       {
+        id: '0b2d1dba-264f-4fb6-b1a3-63c5c40db848',
         provider: {
           identity: '0x1b03b513fba4bd4868edd340f77da0c920150f0a',
           country: 'us'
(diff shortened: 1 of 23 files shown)


src/app/bug-reporting/tequilapi-client-with-metrics.js
```diff
@@ -116,6 +132,132 @@ class TequilapiClientWithMetrics implements TequilapiClient {
   async location (timeout: ?number): Promise<ConsumerLocationDTO> {
     return this._client.location(timeout)
   }
+
+  async sessionsList (): Promise<SessionDto[]> {
```
shroomist: can we do 
```
class TequilapiClientWithMetricsWithFakeSessions extends TequilapiClientWithMetrics {
sessionsList() {...}
}
```
and return it from DI?

src/libraries/unit-converter.js
```diff
@@ -15,46 +15,75 @@
...
- * @param {number} val
+ * @param {number} bytes
  * @returns {{value:number,units:string}} result - holds value and units
  * @throws if argument is null
  */
 
-function bytesReadable (val) {
-  if (typeof val !== 'number') {
+function bytesReadable (bytes: number): Bytes {
+  if (typeof bytes !== 'number') {
```
shroomist: function signature should ensure this. why do we still need this validation?

src/libraries/unit-converter.js
```diff
@@ -15,46 +15,75 @@
...
+  const calculated = size(bytes).calculate('jedec')
   return {
     value: calculated.fixed,
     units: calculated.suffix.replace('i', '')
   }
 }
 
+const bytesReadableDefault: Bytes = { value: '-', units: 'KB' }
+
+function bytesReadableOrDefault (bytes: number): Bytes {
```
shroomist: `function bytesReadable (bytes: number): Bytes {}` should be enough.

src/libraries/unit-converter.js
```diff
@@ -15,46 +15,75 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+// @flow
+
 import size from 'file-size'
 
+type Bytes = {
```
shroomist: I suggest `BytesReadable` for this

src/libraries/unit-converter.js
VaidotasSm: Maybe `bytesReadableOr(bytes, bytesReadableDefault)` to be more explicit. Although I kinda dislike naming, as it's not "verby" and could be confused with values/constructors, but its a matter of taste.

src/renderer/components/session-item.vue
```diff
@@ -0,0 +1,73 @@
...
+    startDate: function () {
+      return getReadableDate(this.dateObject)
+    },
+    startTime: function () {
+      return getReadableTime(this.dateObject)
+    },
+    dateObject: function () {
+      return new Date(this.session.statistics.dateStart)
+    },
+    countryCode: function () {
```
VaidotasSm: Why functions over methods?

shroomist: agreed, you could simplify this to:
```
computed: {
countryCode () {
...
}
```

src/libraries/unit-converter.js
```diff
@@ -15,46 +15,75 @@
...
+  value: string,
```
mipo47: amount

src/renderer/assets/less/4_components/connection-history.less
```diff
@@ -0,0 +1,48 @@
...
+
+.connection-history {
+  table, th, td {
+    border-collapse: collapse;
+    border-spacing: 1px;
+    border: solid 1px @border;
+  }
+
+  table {
+    margin-top: 20px;
```
mipo47: would be nice to move standard margin sizes to variables
to force us use same margins everywhere

src/renderer/components/session-item.vue
```diff
@@ -0,0 +1,73 @@
...
+
+<template>
+  <tr>
+    <td>
+      <CountryFlag :code="countryCode"/>
+      {{ shortIdentity }}
+    </td>
+    <td>{{ startDate }}<br>{{ startTime }}</td>
+    <td>{{ durationString }}</td>
+    <td>⇩{{ received }}<br>⇧{{ sent }}</td>
```
mipo47: Use <br /> instead of <br> because it compatible with XHTML

src/libraries/unit-converter.js
donce: Though about removing these, but these functions are used from *.vue* files, where Flow checks do not work, so it still seems worth keeping for a while.

src/renderer/components/session-item.vue
```diff
@@ -0,0 +1,73 @@
...
+    sent: function () {
+      const sent = bytesReadableOrDefault(this.session.statistics.bytesSent)
+      return sent.value + sent.units
+    },
+    received: function () {
+      const received = bytesReadableOrDefault(this.session.statistics.bytesReceived)
+      return received.value + received.units
+    },
+    shortIdentity: function () {
+      return this.session.proposal.providerId.slice(0, 11) + '...'
```
mipo47:
```
if (this.session.proposal.providerId.length > 11) {
  return this.session.proposal.providerId.slice(0, 11) + '...'
}
return this.session.proposal.providerId
```

src/libraries/unit-converter.js
donce: Prefixed all methods with `format`. @shroomist `bytesReadable` name is already used for function which fails when number is not formattable, instead or returning default.
`bytesReadableOr(bytes, bytesReadableDefault)` looks alright, but I'm a bit hesitant to change - I would have to expose `bytesReadableDefault` then and method usage would be even more complex.

src/libraries/unit-converter.js
```diff
@@ -15,46 +15,75 @@
...
+ * @param {number} bytes
  * @returns {{value:number,units:string}} result - holds value and units
  * @throws if argument is null
  */
 
-function bytesReadable (val) {
-  if (typeof val !== 'number') {
+function formatBytesReadable (bytes: number): BytesReadable {
+  if (typeof bytes !== 'number') {
     throw new Error('provide valid input for conversion')
```
mipo47: throw new Error(`provide valid input for conversion: '${bytes}'`)

src/app/bug-reporting/tequilapi-client-with-metrics.js
donce: Would look much nicer, but I want to remove this mock before merging into master. Added a TODO not to forget about that.

src/renderer/components/session-item.vue
donce: Vue style guide we're using forbids it - it requires `<br>`

src/libraries/unit-converter.js
donce: This is not related to my PR, I haven't touched this :P

shroomist:
some feedback from testers community
> may be good to add the country initials (at least on hover on the flag) as people don't know all the flags

donce:
Waiting for https://github.com/mysteriumnetwork/node/issues/433.

donce:
Created a separate task for country initials hover: https://mysteriumnetwork.atlassian.net/browse/MVPN-301

shroomist:
can we close this?

PR #355, opened 2018-10-17
MVPN-202: fix identity-registration mounting issues

VaidotasSm:
Originally this issue started from missing initial transition, but later investigation showed issues with error thrown from Vuex getter, which was swallowed since Vue doesn't handle such errors by default.

- Added global Vue error logging for such cases.
- Fixed getter to not throw error, but rather empty string and log warning.


src/app/logger.js
```diff
@@ -56,6 +56,10 @@ class Logger {
   error (...data: Array<any>): void {
     this._logger.error(stringifyArgs(data))
   }
+
+  warn (...data: Array<any>): void {
+    this._logger.warn(stringifyArgs(data))
+  }
```
donce: Keeping log levels sorted may improve readability of levels - i.e. `info, warn, error`

src/renderer/dependencies/modules/vue.js
```diff
@@ -46,6 +46,10 @@ function bootstrap (container: Container) {
       }
       Vue.http = Vue.prototype.$http = axios
       Vue.config.productionTip = false
+      Vue.config.errorHandler = (err) => {
+        // eslint-disable-next-line
+        console.error('Unhandled error:', err)
```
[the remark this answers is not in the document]
donce: !!
Yes, I think it should use `logger.error` to sync error message to `main` process as well.
Also, I think this error is not sent to Sentry by default - we should send it using `BugReporter`.

src/renderer/store/modules/identity.js
```diff
@@ -63,7 +64,8 @@ const getters = {
   currentIdentity (state: State): string {
     const identity = state.current
     if (!identity) {
-      throw new Error('Trying to get identity which is not present')
+      logger.warn('Trying to get identity which is not present')
+      return ''
```
donce: !! This error is being tracked in the Sentry currently:

https://sentry.io/mysterium-network/mysterium-vpn/issues/725441001/?environment=production

Just logging it and ignoring it without sending it to sentry will hide the actual problem - we should send it to Sentry using `BugReporter`.

test/unit/renderer/components/identity-registration.spec.js
```diff
@@ -36,10 +36,10 @@ import types from '../../../../src/renderer/store/types'
 describe('IdentityRegistration', () => {
   let rendererCommunication: RendererCommunication
   let mainCommunication: MainCommunication
-  let vue: IdentityRegistration
+  let wrapper: IdentityRegistration
   let store: Vuex.Store
 
-  beforeEach(() => {
+  function mountEverything (stateOverride = {}) {
```
donce: default value of `stateOverride` is unused - we can just delete the default.

test/unit/renderer/components/identity-registration.spec.js
```diff
@@ -69,24 +65,44 @@ describe('IdentityRegistration', () => {
...
     })
 
-    vue = mount(IdentityRegistration, {
+    wrapper = mount(IdentityRegistration, {
       localVue: vm,
       store
     })
+  }
+
+  describe('HTML rendering when State is  not OK', () => {
```
donce: two spaces :⭕️

test/unit/renderer/components/identity-registration.spec.js
```diff
@@ -69,24 +65,44 @@ describe('IdentityRegistration', () => {
...
+    beforeEach(() => {
+      mountEverything({
+        current: null
+      })
+    })
+
+    it('should still render component', () => {
```
donce: When does this null value actually happens? It's probably some sort of bug we should create in Jira?

VaidotasSm:
@donce addressed issues as discussed

src/renderer/dependencies/modules/vue.js
```diff
@@ -46,6 +47,11 @@ function bootstrap (container: Container) {
       }
       Vue.http = Vue.prototype.$http = axios
       Vue.config.productionTip = false
+      Vue.config.errorHandler = (err) => {
+        // eslint-disable-next-line
```
donce: !! No need to ignore eslint anymore.

src/renderer/dependencies/modules/vue.js
```diff
@@ -46,6 +47,9 @@ function bootstrap (container: Container) {
       }
       Vue.http = Vue.prototype.$http = axios
       Vue.config.productionTip = false
+      Vue.config.errorHandler = (err) => {
+        logger.error('Unhandled error:', err)
```
donce: It would be very good for us to know if such errors happen - reporting to bug reporter might helps us detect them faster :)

VaidotasSm:
@donce @shroomist @mipo47 @ignasbernotas I have signed the commits and as a result approves are gone, could you take a look again?

src/renderer/store/modules/identity.js
mipo47: return null
or
return undefined

test/unit/renderer/components/identity-registration.spec.js
shroomist: this happens because we have `<IdentityRegistration>` component in `app.vue`
which depends on unlocked identity.
`<IdentityRegistration>` should be moved to `vpn.vue`

PR #362, opened 2018-10-24
[MVPN-249] Mysterium client upgrading for windows

donce:
- On `master`, upgrading only works for OS X.
- This moves upgrade functionality to `Process.upgrade` and implements it for Windows as well.
- Added more `wait` methods for Monitoring, because waiting for mysterium client to go up/down is required when upgrading.

UPDATE about refactoring:

- Extracted `StatusNotifier` for basic status notifications (only `.onStatus`). `Monitoring`accepts `StatusNotifier`in constructor.
- `TequilapiStatusNotifier` and `MockStatusNotifier` implements `StatusNotifier`, allowing to re-use `Monitoring` class in tests just by switching notifier class.
- Also this will allow to change our monitoring functions without changing monitoring tequilapi notifier.


src/libraries/mysterium-client/monitoring.js
```diff
@@ -64,27 +90,51 @@ class StatusMonitoring {
...
 
+  onNewStatusUp (callback: EmptyCallback): void {
+    this._upSubscriber.subscribe(callback)
+  }
+
+  onStatusChangeUp (callback: EmptyCallback): void {
+    this._changeUpSubscriber.subscribe(callback)
+  }
+
+  waitForNewStatusUpWithTimeout (): Promise<void> {
```
ignasbernotas: !! I don't like the fact that the timeout is hidden within the implementation. I will have no idea when it will timeout without going into the internals. It would be a lot more obvious if the method would accept a timeout argument (ProcessManager could pass it) or these would be passed into the constructor. ProcessManager is the commanding level where it should have control over it.

src/libraries/mysterium-client/monitoring.js
```diff
@@ -64,27 +90,51 @@ class StatusMonitoring {
...
-    this._changeUpSubscriber.subscribe(callback)
+  onNewStatusDown (callback: EmptyCallback): void {
+    this._downSubscriber.subscribe(callback)
   }
 
   onStatusChangeDown (callback: EmptyCallback): void {
     this._changeDownSubscriber.subscribe(callback)
   }
 
+  waitForStatusDownWithTimeout (): Promise<void> {
```
ignasbernotas: I would move these into another class.

src/libraries/mysterium-client/monitoring.js
```diff
@@ -64,27 +90,51 @@ class StatusMonitoring {
...
-    this._downSubscriber.subscribe(callback)
+    this.onNewStatusDown(callback)
     if (this._lastStatus === false) {
       callback()
     }
   }
 
-  onStatusChangeUp (callback: EmptyCallback): void {
-    this._changeUpSubscriber.subscribe(callback)
+  onNewStatusDown (callback: EmptyCallback): void {
```
ignasbernotas: !! this should be a private method and removed from the interface.

src/libraries/mysterium-client/monitoring.js
```diff
@@ -64,27 +90,51 @@ class StatusMonitoring {
...
 
   onStatusUp (callback: EmptyCallback): void {
-    this._upSubscriber.subscribe(callback)
+    this.onNewStatusUp(callback)
     if (this._lastStatus) {
       callback()
     }
   }
 
+  onNewStatusUp (callback: EmptyCallback): void {
```
ignasbernotas: !! this should be a private method and removed from the interface.

src/libraries/mysterium-client/monitoring.js
```diff
@@ -30,20 +32,44 @@ type EmptyCallback = () => void
...
+  onNewStatusDown (callback: EmptyCallback): void,
+  waitForNewStatusDownWithTimeout (): Promise<void>,
+
+  /**
+   * Triggers once service status changes to down.
+   */
+  onStatusChangeDown (callback: EmptyCallback): void
 }
 
 class StatusMonitoring {
```
ignasbernotas: ?? This class does too much. As per my previous comments, I would suggest to restructure it into two classes:
Monitoring class which does all the magic with tequilapi, but doesn't concern itself with business details.
Monitoring.onStatusUp
Monitoring.onStatusDown
Monitoring.onStatusChange

ClientStatusEvents class that 
ClientStatusEvents.waitForStatusUpWithTimeout(timeout?: number)
ClientStatusEvents.waitForStatusDownWithTimeout(timeout?: number)
ClientStatusEvents.other..

ignasbernotas: Also, remember - composition over inheritance, TequilapiMonitoring shouldn't extend StatusMonitoring, instead it should depend on it and implement a slim Monitoring interface.

src/libraries/mysterium-client/monitoring/utils.js
```diff
@@ -0,0 +1,45 @@
...
+  const process = new TequilapiStatusNotifier(tequilapi)
+  const monitoring = new Monitoring(process)
+  const statusUpAsync = promisify(monitoring.onStatusChangeUp.bind(monitoring))
+  monitoring.start()
+
+  return Promise.race([
+    throwErrorAfterTimeout(timeout),
+    statusUpAsync()
+  ]).finally(() => {
+    process.stop()
```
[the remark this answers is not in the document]
shroomist: ?? shouldn't we do `monitoring.stop()` here instead of `process.stop`?

src/app/mysterium-client/process-manager.js
```diff
@@ -137,34 +136,13 @@ class ProcessManager {
 
   async _restartClient () {
```
shroomist: maybe this should now be named `_upgradeClient`?

donce: Yep, updated.

src/libraries/mysterium-client/monitoring/utils.js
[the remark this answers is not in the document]
donce: Yes - `process` name was a left-over from all the refactoring, renamed it to `notifier` and fixed monitoring stopping.

PR #366, opened 2018-11-02
[MVPN-334]: Move identity-registration to vpn page

VaidotasSm:


src/renderer/assets/less/4_components/countries.less
```diff
@@ -28,7 +28,7 @@
     position: absolute;
     top: 0.7rem;
     left: 0.6rem;
-    z-index: 2;
+    z-index: 1;
```
VaidotasSm: Not sure if there was scenario where z-index of 2 was needed? Clicked around and all seems fine, but maybe I missed something.

donce: @shroomist maybe you know this?

src/renderer/app.vue
```diff
@@ -20,8 +20,6 @@
     id="app"
     class="app">
     <div id="content">
-      <IdentityRegistration v-if="paymentsAreEnabled"/>
```
donce: !! `paymentsAreEnabled` are no longer needed, please remove that.

PR #367, opened 2018-11-06
Uninstall mysterium_client service in windows installer/uninstaller

donce:
Some docs: https://www.electron.build/configuration/nsis


shroomist:
looks biutiful, is this all of it?

PR #375, opened 2018-11-14
Fix naming clash by using 'Publisher' name

donce:
Using same name (`subscribers`) for both sides of subscriptions is very confusing - usually they are named `publishers` and `subscribers`.

From wiki:
> In software architecture, publish–subscribe is a messaging pattern where senders of messages, called publishers, do not program the messages to be sent directly to specific receivers, called subscribers, but instead categorize published messages into classes without knowledge of which subscribers, if any, there may be. Similarly, subscribers express interest in one or more classes and only receive messages that are of interest, without knowledge of which publishers, if any, there are.

### DIFF
--- a/src/app/data-fetchers/country-list.js
+++ b/src/app/data-fetchers/country-list.js
@@ -20,7 +20,7 @@ import type { Country } from '../countries'
 import Publisher from '../../libraries/publisher'
 import { getSortedCountryListFromProposals } from '../countries'
 import ProposalDTO from 'mysterium-tequilapi/lib/dto/proposal'
-import type { Callback } from '../../libraries/publisher'
+import type { Subscriber } from '../../libraries/publisher'
 import type { FavoriteProviders } from '../user-settings/user-settings'
 import type { ProposalFetcher } from './proposal-fetcher'
 import type { UserSettingsStore } from '../user-settings/user-settings-store'
@@ -41,8 +41,8 @@ class CountryList {
     this._subscribeToFavoriteChanges()
   }
 
-  onUpdate (listener: Callback<Array<Country>>) {
-    this._publisher.subscribe(listener)
+  onUpdate (listener: Subscriber<Array<Country>>) {
+    this._publisher.addSubscriber(listener)
   }
 
   _subscribeToFavoriteChanges () {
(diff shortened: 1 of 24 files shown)


src/app/data-fetchers/country-list.js
```diff
@@ -42,7 +42,7 @@ class CountryList {
   }
 
   onUpdate (listener: Callback<Array<Country>>) {
-    this._listeners.subscribe(listener)
+    this._publisher.subscribe(listener)
```
shroomist: `this._publisher.subscribe` reads unnatural.
feels like it's the publisher is subscribing to something
```suggestion
    this._publisher.addSubscription(listener)
```

```suggestion
    this._publisher.addListener(listener)
```

src/app/data-fetchers/tequilapi-proposal-fetcher.js
```diff
@@ -64,11 +64,11 @@ class TequilapiProposalFetcher implements ProposalFetcher {
   }
 
   onFetchedProposals (subscriber: Callback<ProposalDTO[]>): void {
-    this._proposalSubscriber.subscribe(subscriber)
+    this._proposalPublisher.subscribe(subscriber)
   }
 
   onFetchingError (subscriber: Callback<Error>): void {
-    this._errorSubscriber.subscribe(subscriber)
+    this._errorPublisher.subscribe(subscriber)
```
shroomist: in some cases you use `Publisher.subscribe(subscriber)` in others `Publisher.subscribe(listener)`
in here I suggest 
```suggestion
    this._errorPublisher.subscribe(handleError)
```

src/app/data-fetchers/country-list.js
donce: Renamed to `.addSubscriber` and `.removeSubscriber`, since subscriber is used most commonly :)

src/app/data-fetchers/tequilapi-proposal-fetcher.js
donce: I want to keep this naming consistent - `handleError` is a bit confusing, because this subscriber can just log an error - I wouldn't call logging an error handling.

donce:
@shroomist addressed the issues.

PR #376, opened 2018-11-16
[MVPN-325] Filter and order proposals/countries

mipo47:


src/app/data-fetchers/tequilapi-proposal-fetcher.js
```diff
@@ -18,24 +18,38 @@
...
 import ProposalDTO from 'mysterium-tequilapi/lib/dto/proposal'
+import ProposalsQuery from 'mysterium-tequilapi/lib/adapters/proposals-query'
 import type { TequilapiClient } from 'mysterium-tequilapi/lib/client'
 import { FunctionLooper } from '../../libraries/function-looper'
 import type { Subscriber } from '../../libraries/publisher'
 import Publisher from '../../libraries/publisher'
 import type { ProposalFetcher } from './proposal-fetcher'
 
+const proposalsQueryWithMetric = new ProposalsQuery({ fetchConnectCounts: true })
+const filterFailedProposals = (proposal: ProposalDTO) => {
```
Waldz: !! This logic does not fit here. It should in view layer (by terms of MVC)

src/app/mysterium-vpn.js
```diff
@@ -342,8 +342,8 @@ class MysteriumVpn {
     const handleProposalFetchError = (error: Error) => {
       logException('Proposal fetching failed', error)
     }
-    this._communication.proposalsUpdate.on(() => {
-      this._proposalFetcher.fetch().catch((err: Error) => {
+    this._communication.proposalsUpdate.on((showMore: boolean) => {
+      this._proposalFetcher.fetch(showMore).catch((err: Error) => {
```
Waldz: To show more is UI component logic

src/app/data-fetchers/tequilapi-proposal-fetcher.js
mipo47: View doesn't know about ProposalDTO and metrics

src/app/mysterium-vpn.js
mipo47: The only way UI can get proposals is communication, so here is transport

src/app/countries/index.js
```diff
@@ -120,6 +129,18 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
   return proposal.serviceDefinition.locationOriginate.country
 }
 
+function getCountrySuccessRateFromProposal (proposal: ProposalDTO): number {
+  if (!proposal.metrics || !proposal.metrics.connectCount) {
```
donce: So if metrics do not come through, all proposals will be hidden?

src/app/countries/index.js
```diff
@@ -120,6 +129,18 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
...
+    return UNKNOWN_PROPOSAL_SUCCESS_RATE
+  }
+  const count = proposal.metrics.connectCount
+  const total = count.success + count.fail
```
donce: ?? What about `count.timeout`? We can handle them just like `count.fail`.

src/app/countries/index.js
```diff
@@ -120,6 +129,18 @@ function getCountryCodeFromProposal (proposal: ProposalDTO): ?string {
...
+  if (total > 0) {
+    return count.success / (count.success + count.fail)
```
donce: You can re-use `total`
```suggestion
    return count.success / total
```

src/app/data-fetchers/tequilapi-proposal-fetcher.js
donce: `filter` is just one use-case - it's hard to understand what `true` and `false` return values mean:
```suggestion
const isProposalSuccessful = (proposal: ProposalDTO): boolean => {
```

src/app/data-fetchers/tequilapi-proposal-fetcher.js
```diff
@@ -18,24 +18,38 @@
...
+  if (proposal.metrics && proposal.metrics.connectCount) {
+    const count = proposal.metrics.connectCount
+    if (count.success < 30 && count.fail > 0) {
+      return false
```
donce: ?? What about timeouts?
Also, don't quite get the logic behind these rules.

src/app/data-fetchers/tequilapi-proposal-fetcher.js
```diff
@@ -51,8 +65,13 @@ class TequilapiProposalFetcher implements ProposalFetcher {
   /**
    * Forces proposals to be fetched without delaying.
    */
-  async fetch (): Promise<ProposalDTO[]> {
-    const proposals = await this._api.findProposals()
+  async fetch (showMore: boolean): Promise<ProposalDTO[]> {
+    let proposals = await this._api.findProposals(proposalsQueryWithMetric)
+
+    this._showMore = showMore
```
donce: !! This looks really buggy, because at the same time `fetch`is called by `FunctionLooper` and by any outside caller - this can get messy :(

We should separate `showMore` setting from `fetch` - that way we can explicitly change `showMore` only 
from outside, where it's needed, not in each iteration of `FunctionLooper`.

src/app/data-fetchers/tequilapi-proposal-fetcher.js
Waldz: I strongly disagree that `ProposalFetcher` is *Model* (service)
And I strongly disagree that it should be coupled with UI element `_showMore` - which belong to *View*

mipo47: Ok. currently moving filtering to UI side, thanks for comment

src/app/mysterium-vpn.js
mipo47: will be fixed

src/app/data-fetchers/tequilapi-proposal-fetcher.js
Waldz: There is middle option - have usecase like this:
```
var fetcher = ProposalFetcher(proposalsFilterFunction)
fetcher.fetch()
```

Passing filter function decouples service from view

src/app/data-fetchers/tequilapi-proposal-fetcher.js
```diff
@@ -18,24 +18,29 @@
...
 import type { ProposalFetcher } from './proposal-fetcher'
 
+const proposalsQueryWithMetric = new ProposalsQuery({ fetchConnectCounts: true })
+
 class TequilapiProposalFetcher implements ProposalFetcher {
   _api: TequilapiClient
   _loop: FunctionLooper
   _proposalPublisher: Publisher<ProposalDTO[]> = new Publisher()
   _errorPublisher: Publisher<Error> = new Publisher()
+  _showMore: boolean = false
```
donce: !!
```suggestion
```

src/app/data-fetchers/tequilapi-proposal-fetcher.js
mipo47: fixed

src/app/mysterium-vpn.js
mipo47: fixed

src/app/countries/index.js
mipo47: currently ignored

src/app/data-fetchers/tequilapi-proposal-fetcher.js
```diff
@@ -52,7 +57,7 @@ class TequilapiProposalFetcher implements ProposalFetcher {
    * Forces proposals to be fetched without delaying.
    */
   async fetch (): Promise<ProposalDTO[]> {
-    const proposals = await this._api.findProposals()
+    let proposals = await this._api.findProposals(proposalsQueryWithMetric)
```
donce:
```suggestion
    const proposals = await this._api.findProposals(proposalsQueryWithMetric)
```

src/app/data-fetchers/tequilapi-proposal-fetcher.js
mipo47: fixed

src/renderer/components/country-select.vue
```diff
@@ -110,6 +118,9 @@ export default {
 
       this.country = selectedCountry
       this.$emit('selected', selectedCountry)
+    },
+    filterToggle () {
```
donce:
```suggestion
    toggleShowMore () {
```

PR #377, opened 2018-11-16
feature/mvpn-323 add setting to toggle all providers

shroomist:


shroomist:
#376 has the solution, closing

PR #382, opened 2018-11-30
fix connection: add argument flags to Mysterium client for both platforms

mipo47:


src/libraries/mysterium-client/launch-daemon/launch-daemon-installer.js
```diff
@@ -66,6 +66,7 @@ class LaunchDaemonInstaller implements Installer {
             <string>${this._config.openVPNBin}</string>
             <string>--tequilapi.port</string>
             <string>${this._config.tequilapiPort}</string>
+            <string>--experiment-promise-check=1</string>
```
donce: ?? Key-value pairs seem to be separated in this config, we should keep it consistant if possible - is that possible for this case?
```suggestion
            <string>--experiment-promise-check</string>
            <string>1</string>
```

shroomist:
@donce we need this like it is here for 0.7.0 to work.
@ignasbernotas lets merge and release.

src/libraries/mysterium-client/launch-daemon/launch-daemon-installer.js
shroomist: unfortunately this doesn't work ;/

PR #396, opened 2019-02-08
[Desktop-389] Quality levels

donce:
screenshot 2019-02-08 at 16 48 02


src/renderer/components/quality-indicator.vue
```diff
@@ -0,0 +1,55 @@
...
+  name: 'QualityIndicator',
+  props: {
+    level: {
+      type: Number,
+      default: null
+    }
+  },
+  computed: {
+    iconName: function () {
+      if (this.level === QualityLevel.LOW) {
```
ignasbernotas: could do 
```
const levels = {[QualityLevel.LOW]: 'low', [QualityLevel.MEDIUM]: 'medium',}
return levels[this.level] || 'unknown' 
```

src/app/countries/parsing.js
```diff
@@ -65,68 +53,59 @@ function countryFavoriteMapper (favorites: FavoriteProviders): (Country) => Coun
...
-  } else if (a.successRate < b.successRate) {
+  }
+  if (!a.isFavorite && b.isFavorite) {
     return 1
-  } else if (b.successRate < a.successRate) {
-    return -1
-  } else if (a.name > b.name) {
+  }
+
+  if (a.quality !== null || b.quality !== null) {
```
ignasbernotas: I would probably extract this into a separate function such as `compareQuality(a, b)`, since `compareCountries` function is getting pretty big.

donce: Yes - actually, `mobile` has the same functionality, so we should extract proposals sorting into `mystevium-vpn-js` (and probably filtering, if we'll need this is desktop soon)

Issue #405, opened 2019-02-07
As a user I want be able to see my session history


ignasbernotas:
As a Consumer
I want to see my session history list
In order to understand which nodes I used before

AC:
* View session history list
* Make a release

Waldz:
This ticket is about merging *connection session history* feature to master

PR #410, opened 2019-02-13
[MVPN-203] Connection history

donce:
46519803-5c05fc80-c882-11e8-98c5-ba3b7f07c333

Fixes: https://github.com/mysteriumnetwork/mysterium-vpn/issues/405


src/renderer/components/session-item.vue
```diff
@@ -0,0 +1,81 @@
...
+    return {
+      dateObject: null
+    }
+  },
+  created: function () {
+    this.dateObject = new Date(this.session.dateStarted)
+  },
+  computed: {
+    durationString () {
+      return formatTimeDisplayOrDefault(this.session.duration)
```
ignasbernotas: ?? Shouldn't this function be moved to time formatter class? Now there are 2 ways to format time and I need to check the internals to figure out which one to use:)

src/renderer/pages/connection-history.vue
```diff
@@ -0,0 +1,69 @@
...
+  name: 'ConnectionHistory',
+  components: { CloseButton, SessionItem },
+  dependencies: ['tequilapiClient'],
+  data: function () {
+    return {
+      sessions: []
+    }
+  },
+  created: function () {
+    this.tequilapiClient.sessionsList().then(sessions => {
```
ignasbernotas: 👍 So nice to see tequilapi calls decoupled from vuex 😄

src/libraries/time-formatter.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  _getDateWithOffset (date: Date): Date {
+    const newDate = new Date(date.getTime())
+    newDate.setMinutes(newDate.getMinutes() - this._minutesOffset)
+    return newDate
+  }
+}
+
+const FORMAT_LOCALE = 'en-GB'
```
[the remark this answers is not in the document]
donce: I think that's what we decided with @AndriusMyst

src/renderer/components/session-item.vue
donce: Extracted separate classes: `TimeFormatter`, `DurationFormatter` and `BytesFormatter`

PR #427, opened 2019-02-18
Simplified tray, tests removed

ignasbernotas:


src/app/mysterium-vpn.js
```diff
@@ -370,9 +371,9 @@ class MysteriumVpn {
 
   _buildTray () {
     logInfo('Building tray')
-    trayFactory(
-      this._communication,
-      this._countryList,
+
+    // need to store the instance, otherwise tray will be garbage collected
```
donce: Why wasn't it garbage collected before?

src/main/tray/factory.js
```diff
@@ -16,42 +16,47 @@
...
-  countryList.onUpdate(countries => tray.setCountries(countries))
+  let theme = 'passive'
+
+  // once we upgrade our Electron version, we can use this:
+  // const { remote } = require('electron')
+  // if (process.platform === 'darwin') {
+  //   const { systemPreferences } = remote
+  //
+  //   theme = systemPreferences.isDarkMode() ? 'passive' : 'active'
+  // }
```
donce: !! We should not add dead code - please create an issue in github or a private reminder for this.

Also, I'm pretty sure we won't need this - dark theme inverts colors of monochrome tray icons, so we can have single tray icon.

src/main/tray/factory.js
```diff
@@ -16,42 +16,47 @@
...
-  }
-
-  const tray = new Tray(trayFactory, templateBuilder, menuBuilder, iconPath)
-  tray.build()
-
-  communication.connectionStatusChanged.on((change: ConnectionStatusChangeDTO) => {
-    tray.setStatus(change.newStatus)
-  })
-  countryList.onUpdate(countries => tray.setCountries(countries))
+  let theme = 'passive'
```
donce:
```suggestion
  const theme = 'passive'
```

src/app/mysterium-vpn.js
```diff
@@ -17,7 +17,7 @@
 
 // @flow
 
-import { app, BrowserWindow } from 'electron'
+import { app, BrowserWindow, Tray } from 'electron'
```
donce: !! Please don't couple this class to `electron` even more - we should introduce some abstraction instead of relying on `electron` in our god class.

I.e. we can have `class TrayBuilder` with `build(): void` - we can keep reference only to builder, and builder will keep reference to tray internally.

src/main/tray/factory.js
```diff
@@ -16,42 +16,47 @@
...
+  //   const { systemPreferences } = remote
+  //
+  //   theme = systemPreferences.isDarkMode() ? 'passive' : 'active'
+  // }
+
+  const icon = `tray-${theme}-mode.png`
+
+  const contextMenu = Menu.buildFromTemplate([
+    {
+      label: 'Show app',
```
donce: ?? Not just to check you connection status, you need to click on tray, select "show app", and then close the window. That is very painful.

We should at least keep `Status: Connected` item, or have some other easier way to see your status.

src/main/tray/translations.js
```diff
@@ -1,31 +0,0 @@
...
- * You should have received a copy of the GNU General Public License
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
- */
-
-const translations = {
-  name: 'MysteriumVpn',
-  connect: 'Connect',
-  disconnect: 'Disconnect',
-  showWindow: 'Show window',
-  toggleDeveloperTools: 'Toggle developer tools',
```
donce: ?? Shouldn't we keep this for development build? This is pretty useful, you're not forced to learn shortcut instantly

PR #433, opened 2019-02-20
[MVPN-430] Fix country not being resolved with some locales

donce:
Not sure why we were using locale here o.O This does not work with some locales:
```
> 'IT'.toLocaleLowerCase('tr')
'ıt'
```
We don't need any locales:
```
> 'IT'.toLowerCase()
'it'
```

### DIFF
--- a/src/app/countries/utils.js
+++ b/src/app/countries/utils.js
@@ -41,8 +41,7 @@ function limitedLengthString (value: string, maxLength: ?number = null): string 
 }
 
 function isCountryKnown (countryCode: ?string): boolean {
-  return countryCode != null &&
-    typeof countries[countryCode.toLowerCase()] !== 'undefined'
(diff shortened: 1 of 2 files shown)


src/app/countries/utils.js
```diff
@@ -42,7 +42,7 @@ function limitedLengthString (value: string, maxLength: ?number = null): string
 
 function isCountryKnown (countryCode: ?string): boolean {
   return countryCode != null &&
-    typeof countries[countryCode.toLocaleLowerCase()] !== 'undefined'
+    typeof countries[countryCode.toLowerCase()] !== 'undefined'
```
Waldz: Why do need to compare to such strings? I am sure API never returns 'undefined'

vkuznecovas: doesn't good old `return !!countryCode` not work?

donce: Fixed

PR #449, opened 2019-03-18
Service stopping and app quit fix

ignasbernotas:
Apparently prior to this fix, the mac/windows apps wouldn't wait for the cleanup tequilapi calls to finish. This fixes forces the myst process to quit in OSX and stops running services in windows.

### DIFF
--- a/src/app/mysterium-vpn.js
+++ b/src/app/mysterium-vpn.js
@@ -125,7 +125,13 @@ class MysteriumVpn {
       // this is needed to wait for async ops to finish
       event.preventDefault()
 
-      await this.onWillQuit()
+      try {
+        await this.onWillQuit()
+      } catch (e) {
+        this._bugReporter.captureErrorException(e)
+      }
+
+      process.exit(1)
     })
     // fired when app activated
     app.on('activate', () => {
@@ -295,8 +301,6 @@ class MysteriumVpn {
       logException('Failed to stop process', e)
       this._bugReporter.captureErrorException(e)
     }
-
-    process.exit(1)
   }
 
   // make sure terms are up to date and accepted
--- a/src/app/mysterium-vpn.js
+++ b/src/app/mysterium-vpn.js
@@ -121,7 +121,12 @@ class MysteriumVpn {
     // fired when all windows are closed
     app.on('window-all-closed', () => this.onWindowsClosed())
     // fired just before quitting, this should quit
-    app.on('will-quit', () => this.onWillQuit())
+    app.on('will-quit', async (event) => {
+      // this is needed to wait for async ops to finish
+      event.preventDefault()
+
+      await this.onWillQuit()
+    })
     // fired when app activated
     app.on('activate', () => {
       try {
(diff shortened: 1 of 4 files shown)


src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -82,13 +82,33 @@ class ServiceManagerProcess implements Process {
...
       await this._tequilapi.connectionCancel()
+      await this.stopServices()
     } catch (err) {
       if (err.name !== TequilapiError.name) {
         throw err
       }
     }
   }
 
+  async stopServices (): Promise<void> {
```
donce: Stopping all services looks quite generic - maybe we should add this to `ProviderService` to be able to re-use this?

src/app/mysterium-vpn.js
```diff
@@ -121,7 +121,12 @@ class MysteriumVpn {
     // fired when all windows are closed
     app.on('window-all-closed', () => this.onWindowsClosed())
     // fired just before quitting, this should quit
-    app.on('will-quit', () => this.onWillQuit())
+    app.on('will-quit', async (event) => {
+      // this is needed to wait for async ops to finish
+      event.preventDefault()
+
+      await this.onWillQuit()
```
donce: ??
- What happens if `onWillQuit` doesn't finish - app is unable to quit? In that case, shouldn't we add some timeout?
- What happens if I press Cmd+Q 2 times - two parallel stopping request sequences will start?

src/libraries/mysterium-client/service-manager/service-manager-process.js
```diff
@@ -82,13 +82,33 @@ class ServiceManagerProcess implements Process {
...
+    let services = []
+    try {
+      services = await this._tequilapi.serviceList()
+    } catch (err) {
+      if (err.name !== TequilapiError.name) {
```
Waldz: Code is quite unreadable why `TequilapiError` case is so special

src/libraries/mysterium-client/service-manager/service-manager-process.js
ignasbernotas: Will move this once it's needed:)

src/libraries/mysterium-client/service-manager/service-manager-process.js
ignasbernotas: Not sure, I don't think this can happen anyway (unless there's a syntax error or something), I just copied it from above. @donce any ideas?

Waldz: lets not add code we dont need then

src/app/mysterium-vpn.js
ignasbernotas: 1. Wrapped it in try catch (even though everything in `onWillQuit` is wrapped in `try` blocks) and moved process.exit(1) into this function for clarity.
2. Yes. It will execute the function multiple times, but it doesn't matter. Once the first one is finished `process.exit(1)` will kill it not waiting for others to finish.

src/libraries/mysterium-client/service-manager/service-manager-process.js
ignasbernotas: Removed

PR #459, opened 2019-03-20
Feature/tab navigation modals

ignasbernotas:
Screenshot 2019-03-20 at 16 52 17
Screenshot 2019-03-20 at 16 51 57


donce:
If I start providing service and switch tabs while service is starting, I end up in VPN page with service running - maybe we should cancel service starting in this case?

ignasbernotas:
@donce Fixed CI, added addition in-between state checking for displaying the modal before navigation:)

src/renderer/pages/provider.vue
```diff
@@ -135,7 +143,7 @@ export default {
       }
     },
     buttonText () {
-      if (this.pendingRequest) {
+      if (this.pendingStartRequest || this.pendingStopRequest) {
         return 'Please wait...'
```
donce: ?? Why do we have "Starting.." message few lines below, if we don't actually show it when `pendingStartRequest === true`?
Maybe:
- in case of `pendingStartRequest`, we should show starting message
- in case of `pendingStopRequest`, we should show stopping message

src/renderer/pages/provider.vue
```diff
@@ -205,13 +213,33 @@ export default {
...
+      if (page !== 'vpn') {
+        return
+      }
+
+      if (this.status !== ServiceStatus.NOT_RUNNING || this.pendingStartRequest) {
+        this.showTabModal = true
+        return
+      }
+
+      this.gotToVpn()
```
donce: ?? If `this.pendingStopRequest === true`, then we will go to VPN page before it is actually stopped.
`toVpnPage` acts differently - it waits for service to stop and only when switches to VPN.

Can we unify these 2 paths - either both wait for service to stop, or non of them does that?
Waiting for service to stop seems more stable aproach, but less UX friendly - if we switch to VPN page instantly in both cases, UX is better, but than we need to handle service not being stopped in VPN page.

src/renderer/pages/provider.vue
```diff
@@ -205,13 +213,33 @@ export default {
...
+    },
+    gotToVpn () {
```
donce:
```suggestion
    goToVpn () {
```

src/renderer/pages/vpn.vue
```diff
@@ -158,12 +171,40 @@ export default {
...
+      }
 
       this.countryList = countries
+    },
+    onTabClick (page) {
+      if (page !== 'provider') {
+        return
+      }
+
+      if (this.statusCode > -1) {
```
donce: Why do we need `statusCode`? It's very hard to understand the intent. We already have `ConnectionStatus` enum somewhere for that.

```suggestion
      if (this.status != ConnectionStatus.NotConnected) {
```

src/renderer/assets/less/5_partials/modal.less
```diff
@@ -56,4 +69,11 @@
   &__action {
     margin: 0 0.5rem;
   }
+  &-button {
```
donce: `__` is used for elements (http://getbem.com/naming/)
```suggestion
  &__button {
```

src/renderer/assets/less/5_partials/modal.less
```diff
@@ -18,12 +18,25 @@
...
+      border: 1px solid @mainLight;
+      border-radius: 4px;
+      background-color: @white;
+      .modal__actions {
+        display: block;
+        text-align: center;
+        padding-bottom: 0;
+      }
+    }
+    &--s {
```
donce: It's hard to understand the meaning of `s`
```suggestion
    &--small {
```

src/renderer/pages/vpn.vue
```diff
@@ -158,12 +171,40 @@ export default {
...
+        this.showTabModal = true
+        return
+      }
+
+      this.goToServicePage()
+    },
+    toServicePage () {
+      if (this.statusCode > -1) {
+        this.$store.dispatch(type.DISCONNECT)
```
donce: Shouldn't we await here?
```suggestion
        await this.$store.dispatch(type.DISCONNECT)
```

src/renderer/pages/vpn.vue
```diff
@@ -158,12 +171,40 @@ export default {
...
+
+      this.goToServicePage()
+    },
+    goToServicePage () {
+      this.$router.push('/provider')
     }
   },
   async mounted () {
+    if (await this.providerService.isActive()) {
+      return this.goToServicePage()
```
donce: !! We shouldn't just return here some of `mounted` actions below have to be executed even if app is launched with active provider, i.e. `sendAppStartSuccessEvent`.

test/unit/renderer/pages/vpn.spec.js
```diff
@@ -81,7 +88,7 @@ describe('Vpn', () => {
   }
 
   describe('when payments are enabled', () => {
-    beforeEach(() => {
+    beforeEach(async () => {
```
donce: Why? :O
```suggestion
    beforeEach(() => {
```

src/renderer/pages/provider.vue
ignasbernotas: `this.pendingStartRequest` and `this.status` are not related, we only need to make sure that during the start/stop no button actions can be executed. Status can change independently from that.

test/unit/renderer/pages/vpn.spec.js
ignasbernotas: Async operations in mounted don't get executed in time for tests unless this has async

src/renderer/pages/provider.vue
ignasbernotas: Unified to await :)

src/renderer/pages/vpn.vue
ignasbernotas: changed

ignasbernotas:
@donce Fixed major issues regarding async/await, added explanations for questions. Lets merge this mofo already :)

src/renderer/pages/provider.vue
donce: Yes, but we shouldn't display two different string for the same thing - now, both "Connecting" and "Please wait..." will be displayed for connecting state, depending on how node will go into connecting state. We should unify these - i.e. always show "Connecting" in this case.

src/renderer/pages/provider.vue
```diff
@@ -205,13 +213,33 @@ export default {
...
+      this.pendingStopRequest = false
     },
 
     onStatusChange (newStatus) {
       this.status = newStatus
       // TODO: show error if status changes from "Starting" to "NotRunning"
       // TODO: show error if service ends unexpectedly, without stoping service
+    },
+    async stopAndGoToVpn () {
+      await this.providerService.stop()
```
donce: ?? This can fail, and nobody will know about it - not us, not even user.
We should track this failure as a warning in Sentry and we should show some feedback to user, why he are unable to switch to VPN tab.

src/renderer/assets/less/5_partials/modal.less
donce: `s` in not so self-explanatory for `height: auto`, could we have `&--small`?

src/renderer/assets/less/5_partials/modal.less
donce: !!

src/renderer/pages/provider.vue
donce: !!

src/renderer/pages/provider.vue
```diff
@@ -205,13 +213,33 @@ export default {
...
+
+      this.goToVpn()
+    },
+    async onTabClick (page) {
+      if (page !== 'vpn') {
+        return
+      }
+
+      if (this.status !== ServiceStatus.NOT_RUNNING || this.pendingStartRequest) {
```
donce: ?? Why are we not checking `pendingStopRequest`?
```suggestion
      if (this.status !== ServiceStatus.NOT_RUNNING || this.pendingStartRequest || this.pendingStopRequest) {
```

src/renderer/pages/vpn.vue
```diff
@@ -158,13 +162,42 @@ export default {
...
+      if (this.status !== STATUS_NOT_CONNECTED) {
+        this.showTabModal = true
+        return
+      }
+
+      this.goToProviderPage()
+    },
+    async stopAndGoToProviderPage () {
+      if (this.status !== STATUS_NOT_CONNECTED) {
+        await this.$store.dispatch(type.DISCONNECT)
```
donce: ?? This can fail as well, resulting in no feedback to user request to switch tab, and this probably won't be visible in Sentry - can we fix this?

test/unit/renderer/pages/vpn.spec.js
donce: Ok, this behaviour is very strange..

src/renderer/pages/provider.vue
```diff
@@ -222,7 +223,16 @@ export default {
       // TODO: show error if service ends unexpectedly, without stoping service
     },
     async stopAndGoToVpn () {
-      await this.providerService.stop()
+      try {
+        await this.providerService.stop()
+      } catch (e) {
+        const message = 'Failed to stop service during navigation:' + e.message
```
donce: This is perfect for sentry, but we shouldn't show technical json errors for user - they are usually in golang format, i.e. `server error: request failed with ...`

src/renderer/pages/vpn.vue
```diff
@@ -173,16 +176,25 @@ export default {
...
       this.goToProviderPage()
     },
     async stopAndGoToProviderPage () {
-      if (this.status !== STATUS_NOT_CONNECTED) {
-        await this.$store.dispatch(type.DISCONNECT)
+      if (!this.isNotConnected) {
+        try {
+          await this.$store.dispatch(type.DISCONNECT)
+        } catch (e) {
+          const message = 'Failed to stop VPN during navigation:' + e.message
```
donce: Same as before - we shouldn't show tequilapi error message in user error.

Issue #465, opened 2019-03-26
Show service status in Tray when service is started


ignasbernotas:
As a user

I want to be able to check if my service is running via Tray menu

AC:
- Show "Service: running", "Service: stopped" in tray menu

PR #469, opened 2019-03-27
Show service status in system tray

tadaskay:
Ref: https://github.com/mysteriumnetwork/mysterium-vpn/issues/465

- Added provider service status menu item in the system tray
- Renamed `Status` to `VPN` now that we have 2 statuses (VPN connection and Provider service)

Screenshot 2019-03-27 at 11 05 08


src/app/communication/main-communication.js
```diff
@@ -24,7 +24,7 @@ import type {
   ConnectionStatusChangeDTO,
   CountriesDTO,
   CurrentIdentityChangeDTO,
-  FavoriteProviderDTO,
+  FavoriteProviderDTO, ProviderServiceStatusChangeDTO,
```
donce:
```suggestion
  FavoriteProviderDTO,
  ProviderServiceStatusChangeDTO,
```

src/app/communication/dto.js
```diff
@@ -61,3 +62,7 @@ export type MetricDto = {
   metric: Metric,
   value: mixed
 }
+
+export type ProviderServiceStatusChangeDTO = {
```
donce: I don't think we really need this DTO - we can use `ServiceStatus` directly.

src/app/communication/message-transport.js
```diff
@@ -23,7 +23,7 @@ import type {
   ConnectionStatusChangeDTO,
   CountriesDTO,
   CurrentIdentityChangeDTO,
-  FavoriteProviderDTO,
+  FavoriteProviderDTO, ProviderServiceStatusChangeDTO,
```
donce:
```suggestion
  FavoriteProviderDTO,
  ProviderServiceStatusChangeDTO,
```

tadaskay: Fixed, thanks!

src/app/communication/dto.js
tadaskay: Removed, thanks!

src/app/communication/main-communication.js
tadaskay: 👍

test/unit/main/tray/menu-builder.spec.js
```diff
@@ -85,44 +85,44 @@ describe('tray', () => {
...
       it('renders menu items without disconnect when not connected', () => {
         const items = builder.updateConnectionStatus(ConnectionStatus.NOT_CONNECTED).build()
-        expect(items[1].type).to.equal(separator)
-        expect(items[2].label).to.equal(translations.connect)
-        expect(items[3].type).to.equal(separator)
-        expect(items[4].label).to.equal(translations.showWindow)
-        expect(items[5].label).to.equal(translations.toggleDeveloperTools)
-        expect(items[6].type).to.equal(separator)
-        expect(items[7].label).to.equal(translations.quit)
+        expect(items[2].type).to.equal(separator)
```
donce: Maybe it makes sense to test 1st element as well?

tadaskay: Sorry, I missed this coment. What do you think about testing independently from absolute order?

donce: That's alright, that was just a small comment. What do you mean by testing independently from absolute order?

tadaskay: If there's no need to be so strict about the order of the elements then you could test that menu contains such item, not necessarily at the index `2`. That way you ensure that the functionality is there and at the same time make the test future-proof by not failing when adding new items to the list.

Issue #458, opened 2019-03-19
Allow entering ETH address for Bounty Program


donce:

donce:
@AndriusMyst should we give some feedback if user enters invalid address? I.e.
As a user
When I enter incorrect ETH address
I want to know that I need to fix it

donce:
Decided we don't need any feedback about bad input - entering bad or empty value end pressing OK will result in no feedback to user.

mjankauskas:
I hear about this task in standups for few days already, but task is still in backlog @donce is this the right status of it?

donce:
Basically, yes - related PRs tagged in this issue shows the status.

PR #471, opened 2019-03-27
[MVPN-458] Eth address for Bounty

donce:
For: https://github.com/mysteriumnetwork/mysterium-vpn/issues/458

Simple functionality to allow entering it, without the ability to update it or any validation.

Screenshot 2019-03-27 at 11 55 42
Screenshot 2019-03-27 at 11 56 05
Screenshot 2019-03-27 at 12 08 44


src/renderer/assets/less/4_components/identity-registration.less
```diff
@@ -50,11 +50,6 @@
   }
 }
 
-.registration-instructions-top-row {
```
tadaskay: It would be great to provide more context in the commit message rather than vague _Refactor_, when possible. I can't really figure out what's being changed here.

src/renderer/assets/less/4_components/identity-menu.less
```diff
@@ -117,3 +58,22 @@
     border-radius: 5px;
   }
 }
+
+.consumer-id-view {
+  margin-top: 2em;
```
tadaskay: I see that you are only moving things around, however, there's a lot of mixing between `px/em/rem`. Why not just use `rem`?

donce: Unified to rem

PR #496, opened 2019-04-18
Use static 1194 openvpn port

### LINKED ISSUE #495: Use 1194 openvpn port in desktop client
In the new client, we can probably add randomization option. For now, let's send `--openvpn.port 1194` to revert to previous behavior.

tadaskay:
Fixes #495 

SO that people can configure NAT port forwarding for a known port number
if uPNP doesn't work for them

### DIFF
--- a/src/libraries/mysterium-client/service-manager/service-manager-installer.js
+++ b/src/libraries/mysterium-client/service-manager/service-manager-installer.js
@@ -152,6 +152,7 @@ class ServiceManagerInstaller implements Installer {
         `--data-dir=${this._config.dataDir}`,
         `--runtime-dir=${this._config.runtimeDir}`,
         `--openvpn.binary=${this._config.openVPNBin}`,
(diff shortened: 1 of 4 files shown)


src/libraries/mysterium-client/service-manager/service-manager-installer.js
```diff
@@ -152,6 +152,7 @@ class ServiceManagerInstaller implements Installer {
         `--data-dir=${this._config.dataDir}`,
         `--runtime-dir=${this._config.runtimeDir}`,
         `--openvpn.binary=${this._config.openVPNBin}`,
+        `--openvpn.port=1194`,
```
donce: Is it a temporary solution? If not, we should extract it to config to keep it consistant.

tadaskay: Yeah, it will be superseeded by `dvpn-desktop`

